Issue #14, opened 2018-05-09
DNSPlugin for Cloudflare


rian-hout:
I wrote a DNSPlugin for Cloudflare.  I'm having trouble make a pull request to upload the file.  I'm receiving a permissions error.  Can you help with this?

rmbolger:
Did you fork this project into your own GitHub profile and make changes to the fork? Or did you just clone this project directly and make changes there?

If it was the latter, the problem is that you don't have access to push changes directly into this repository. Here's a decent guide on contributing to a project on GitHub.
https://akrabat.com/the-beginners-guide-to-contributing-to-a-github-project/

If you're getting a permissions error uploading to your own fork, I have no clue what might be wrong. You might need to contact GitHub support.

rian-hout:
I just did a fork and created a pull request.

rmbolger:
This is now in the master branch and will be included in the next release.

PR #15, opened 2018-05-09
Cloudflare DNS Plugin

rian-hout:
Added plugin for Cloudflare


rmbolger:
Awesome. Thanks again for this. This addresses issue #14 

Give me some time to review it. I'm at a conference this week and don't have a ton of spare time. This is also the first pull request I've ever had in a project, so I'm still new to how it all works.

Can you possibly add a Cloudflare-Readme.md to go with this for potentially new Cloudflare users who might not be familiar with setting up API access?  Most of the other plugins have one now that you can use for reference. You shouldn't need to re-submit the pull request though. I could be wrong, but I think additional commits on your fork will get added to this one.

rian-hout:
I've added a readme.

rmbolger:
I commented out references to $Domain so they don't pollute the pipeline (which it looks like you probably intended to do but forgot). I also tweaked the readme text a bit. Otherwise, this looks good. Thanks again.

PR #17, opened 2018-05-11
Update Azure.ps1

juliansiebert:
Hi,
first off, thank you for this!

Don't know where the "$pargs" variable came from :)

Best regards
Julian


rmbolger:
Whoops. A bit of console test code that ended up in the final file. Thanks for catching this.

PR #42, opened 2018-06-04
Add support for provided Azure access token and IMDS

perbergland:
Instead of passing in a tenant and credentials, you can instead pass in a token that you can get from various places, e.g. the MSI local endpoint or the Azure-CLI 2.0 "az account get-access-token --resource xxx" command.

### DIFF
--- a/Posh-ACME/DnsPlugins/Azure-Readme.md
+++ b/Posh-ACME/DnsPlugins/Azure-Readme.md
@@ -41,7 +41,7 @@ $roleDef = Get-AzureRmRoleDefinition -Name "DNS Zone Contributor"
 $roleDef.Id = $null
 $roleDef.Name = "DNS TXT Contributor"
 $roleDef.Description = "Manage DNS TXT records only."
-$roleDef.Actions.RemoveRange(0,$role.Actions.Count)
+$roleDef.Actions.RemoveRange(0,$roleDef.Actions.Count)
 $roleDef.Actions.Add("Microsoft.Network/dnsZones/TXT/*")
 $roleDef.Actions.Add("Microsoft.Network/dnsZones/read")
 $roleDef.Actions.Add("Microsoft.Authorization/*/read")
--- a/Posh-ACME/DnsPlugins/Azure-Readme.md
+++ b/Posh-ACME/DnsPlugins/Azure-Readme.md
@@ -1,18 +1,18 @@
 # How To Use the Azure DNS Plugin
 
-This plugin works against the [Azure DNS](https://azure.microsoft.com/en-us/services/dns/) provider. It is assumed that you already have an active subscription with at least one DNS zone, associated Resource Group, and an account with access to create roles and app registrations. The commands used in this guide will also make use of the [AzureRM.Profile](https://www.powershellgallery.com/packages/AzureRM.profile/5.0.1), [AzureRM.Resources](https://www.powershellgallery.com/packages/AzureRM.Resources/6.0.0), and [AzureRM.Dns](https://www.powershellgallery.com/packages/AzureRM.Dns/5.0.0) modules. But they are not required to use the plugin normally.
-
-**This plugin currently does not work on non-Windows OSes in PowerShell Core. [Click here](https://github.com/rmbolger/Posh-ACME/wiki/List-of-Supported-DNS-Providers) for details.**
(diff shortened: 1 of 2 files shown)


perbergland:
I experimented with parameter sets as you suggested but in the end I found that they were not a good experience - the error reporting was very unhelpful compared to the custom parameter validation and passing on the parameters from one function (Add-DnsTxtAzure) to the one (Connect-AZTenant) where the validation and branching depending on parameter set actually happened was cumbersome.

rmbolger:
Ah yes. The parameter set resolution error can be addressed by setting a default parameter set in the `[CmdletBinding()]` attribute. I should have put that in my original example.

```powershell
function Add-DnsTxtAzure {
    [CmdletBinding(DefaultParameterSetName='SvcPrincipal')]
```

Passing the parameters to sub-functions can be made easier with splatting which is the same thing the main functions do. You add an extra parameter at the end that has `[Parameter(ValueFromRemainingArguments)]` attached to it. Then, instead of calling the function with explicit arguments, you can just splat the calling function's parameters like this.

```powershell
Connect-AZTenant @PSBoundParameters
```

It will pick up and use the parameters it cares about and just store the rest in the extra parameter rather than throwing an error because it doesn't recognize them.

Aside from needing less code for parameter validation, the other reason I like parameter sets so much is that the built-in help makes it way more obvious at a glance that there are two different ways to call the function and which parameters are mandatory. Consider the current syntax help output for `Add-DnsTxtAzure` which makes it look like AZTenantId, AZAppCred, and AZProvidedToken are all optional.

```
SYNTAX
    Add-DnsTxtAzure [-RecordName] <String> [-TxtValue] <String> [-AZSubscriptionId] <String> [[-AZTenantId] <String>]
    [[-AZAppCred] <PSCredential>] [[-AZProvidedToken] <String>] [-ExtraParams <Object>] [<CommonParameters>]
```

Compare that to the syntax when using parameter sets which shows two different calling methods where AZTentantId and AZAppCred are both mandatory in the first one and AZProvidedToken is mandatory in the second.

```
SYNTAX
    Add-DnsTxtAzure [-RecordName] <String> [-TxtValue] <String> [-AZSubscriptionId] <String> [-AZTenantId] <String>
    [-AZAppCred] <PSCredential> [-ExtraParams <Object>] [<CommonParameters>]

    Add-DnsTxtAzure [-RecordName] <String> [-TxtValue] <String> [-AZSubscriptionId] <String> [-AZProvidedToken]
    <String> [-ExtraParams <Object>] [<CommonParameters>]
```

In any case. Give me some time to review. I may not be able to get to it today though. I'll probably try to throw together a version with parameter sets as well just to see what you think.

rmbolger:
Ok, I started digging into this today. Quick question. If the AZProvidedToken value is just an existing access token (JWT) for an already authenticated account, doesn't that mean it will expire not long after the first use (thus breaking the possibility of using it for later renewals)? Or do the Managed Service Identities have access tokens that don't expire?

perbergland:
The MSI tokens seem to have a ttl of an hour just like other azure tokens so that’s why it’s nice to check the expiry date in the tool.
The [AppAuthentication](https://github.com/Azure/azure-sdk-for-net/blob/psSdkJson6/src/SdkCommon/AppAuthentication/README.md) Package in the Azure .Net Sdk has a cache design because of that.

rmbolger:
I'm confused then. If the token expires in an hour, it's basically only good once (or however many certs you generate before it expires). And it can't be used for renewal later. So is the idea that you'll always call `New-PACertificate` with a new copy of the token when you need to renew? Or are the VMs/Services you'd be spinning up short lived and not need a renewal?

After reading up [about MSIs](https://docs.microsoft.com/en-us/azure/active-directory/managed-service-identity/overview) a bit more, they sound like Managed Service Accounts in AD where the idea is that you don't know the password, you can just *use* the account. In Azure's case, it looks like the Instance Metadata Service (http://169.254.169.254/metadata/identity/oauth2/token) is how you would request a current token for an MSI from within the VM. So would it make more sense to have a parameter set that allows fetching a fresh token from IMSD rather than passing an existing token that will expire?

rmbolger:
Ok, so I've been playing around a bit with MSIs and access tokens in Azure. What if in addition to supporting the explicit token, we also support a switch that basically just uses the Azure Instance Metadata Service (IMDS) endpoint directly?  Like this:

https://docs.microsoft.com/en-us/azure/active-directory/managed-service-identity/how-to-use-vm-token#get-a-token-using-azure-powershell

rmbolger:
What do you think of these changes? The existing access token method is still there, renamed to `AZAccessToken`. And there's a third option for MSIs that just has a switch called `AZUseIMDS`.

perbergland:
Looks great! Can't wait to try it out the IMDS version once I carve out some time to set up a minimal VM with this. This would work on a linux VM, right?

rmbolger:
I've only tested from a Windows instance so far. But it should.

Issue #50, opened 2018-06-22
GoDaddy plugin breaks with records in sub-domains


davehope:
I am attempting to issue a certificate for acmetest.int.hope.mx , using the GoDaddy plugin.  I'm doing this as follows:

` New-PACertificate 'acmetest.int.hope.mx' -AcceptTOS -Contact someone@someone.com -DnsPlugin GoDaddy -PluginArgs @{GDKey='xxx'; GDSecret='xxx'} -Debug`

This results in an error:

```
.. trimmed before here ..
DEBUG: Calling GoDaddy plugin to add _acme-challenge.acmetest.int.hope.mx TXT with value NH-jufRQ57OPCLCgTTSHPpu8P1s80NswgATY8ozKt7U
DEBUG: Checking acmetest.int.hope.mx
DEBUG: Checking int.hope.mx
DEBUG: Checking hope.mx
DEBUG: ""

Invoke-RestMethod : {"code":"INACTIVE_DOMAIN","message":"The given domain is not eligible to have its nameservers changed"}
At C:\Program Files\WindowsPowerShell\Modules\Posh-ACME\2.4.0\DnsPlugins\GoDaddy.ps1:125 char:17
+     $response = Invoke-RestMethod -Uri "$apiRoot/$zone/records" `
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (System.Net.HttpWebRequest:HttpWebRequest) [Invoke-RestMethod], WebException
    + FullyQualifiedErrorId : WebCmdletWebResponseException,Microsoft.PowerShell.Commands.InvokeRestMethodCommand
```
PR incoming.

rmbolger:
Good catch and thanks for the PR as well! I'll follow up there.

Dule011:
fix seems to be needed in Add-DnsTxtGoDaddy and Remove-DnsTxtGoDaddy functions. Replacing lines 25 and 107: $name = ($RecordName -split "\.")[0]
with: $name = $RecordName.Split(".")[0..($RecordName.Split(".").count - 3)] -join "."
worked for me.

rmbolger:
@Dule011 I think this is more or less what @davehope's pull request #51 does except instead of making `$name` the first X pieces of the FQDN, it removes the zone name returned by `Find-GDZone`.

Dule011:
[...]

PR #51, opened 2018-06-22
Fix GoDaddy for records outside of apex domain

davehope:
Fix issue  #50 where the record may not be at the top of the zone.

### DIFF
--- a/Posh-ACME/DnsPlugins/GoDaddy.ps1
+++ b/Posh-ACME/DnsPlugins/GoDaddy.ps1
@@ -22,8 +22,9 @@ function Add-DnsTxtGoDaddy {
 
     $headers = @{Authorization = "sso-key $($GDKey):$($GDSecret)"}
 
-    $name = ($RecordName -split "\.")[0]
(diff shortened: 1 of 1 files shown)


davehope:
Some caveats:
Could use more testing, have only tested with hope.mx , acmetest.int.hope.mx and acmetest2.int.hope.mx
Have only spent ~10mins looking through the code, so apologies if I've missed something.

rmbolger:
Thanks again for the PR! Unfortuantely, I can't test this myself at the moment because I don't have any zones hosted with them. The plugin was originally submitted by another user. But let me clarify my understanding. 

The old code incorrectly assumed that the record being written would always live directly in the zone returned by `Find-GDZone`. So in your bug example, it was trying to create a `_acme-challenge` record within `hope.mx` when it should have been using `_acme-challenge.acmetest.int` as the record name for the API call?

Also, does GoDaddy even support explicit sub-zones? Or does everything just live in the purchased domain's zone?

rmbolger:
Nah, this is fine. Their API definitely seems to revolve around the apex domain that was purchased.

PR #54, opened 2018-06-27
Add Dynu DNS provider

alexzorin:
Only tested on PowerShell Core 6.1 preview (all I have access to on Ubuntu Bionic), would appreciate a test run from Windows/lower versions.

LMK if you want my Dynu credentials for that purpose.

### DIFF
--- a/Posh-ACME/DnsPlugins/Dynu-Readme.md
+++ b/Posh-ACME/DnsPlugins/Dynu-Readme.md
@@ -1,0 +1,18 @@
+# How To Use the Dyunu DNS Plugin
+
+This plugin works against the [Dyunu](https://www.dynu.com) DNS provider. It is assumed that you have already setup an account and created the DNS zone(s) you will be working against.
+
+**Please note**: You must bring your own domain. It is not possible to set TXT records on the domains provided by Dynu.
+
+## Setup
+
+You will need to retrieve the Client ID and Secret that will be used to update DNS records. You can find these on your [Dynu Control Panel](https://www.dynu.com/en-US/ControlPanel/APICredentials).
+
+## Using the Plugin
+
+You will need the account's Client ID and Secret to be set as `DyunuClientID` and `DynuSecret`.
+
+```powershell
+$DynuParams = @{DynuClientID='xxxxxxxx'; DynuSecret='xxxxxxxx'}
+New-PACertificate '*.test.example.com','test.example.com' -DnsPlugin Dynu -PluginArgs $DynuParams
+```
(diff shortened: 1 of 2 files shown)


rmbolger:
Hey thanks for this! Looks like I can probably test by creating a free account. So no worries on sharing your credentials.

rmbolger:
Testing looks good so far on Windows PowerShell 5.1. A couple minor things though.

The Dynu API seems to return HTTP 200 responses even when it's responding with an error which means the try/catch blocks for Invoke-RestMethod won't throw on API problems (still good for other problems though). The most likely error people would probably run into is the "Quota Exception" error on free accounts which are limited to 4 records. Right now, the Add function returns as if it succeeded but the record doesn't get created. So we should probably check the response for that and throw if found. For reference, the raw JSON response of that error looks like this:

```javascript
{"type":"Quota Exception","message":"DNS record limit has been reached."}
```

Also, can we rename `$script:AccessToken` to `$script:DynuToken` or `$script:DynuAccessToken` just to keep the module's variable namespace a bit more organized?

Lastly, the OAuth bearer token appears to be good for 8 hours. While it's highly unlikely a single cert request would exceed that timeout unless someone has a huge `-DnsSleep` value, it's remotely possible someone could leave their PowerShell session open for longer than the token is good for and basically get access denied until they open a new session. We might want to track the expiration and auto renew the token if it expires. The Azure plugin does this if you want inspiration. Realistically, I'm also ok just leaving things as-is and fixing it later if someone complains.

alexzorin:
> Also, can we rename $script:AccessToken to $script:DynuToken or $script:DynuAccessToken just to keep the module's variable namespace a bit more organized?

Renamed. Sorry, this is my first attempt at Powershell, didn't even realize that `$script` was global.

> We might want to track the expiration and auto renew the token if it expires

Again, my mind hasn't adapted to PowerShell: I assumed the lifetime of this data was limited to the `New-PACertificate` call. I've added expiration tracking to the re-use.

> The Dynu API seems to return HTTP 200 responses

Yikes. I did not encounter any of these when testing the "bad path". Nice pickup, I've added validation to both Add and Remove.

rmbolger:
This is great for a first attempt at PowerShell! But yeah, variable scoping definitely takes a bit to wrap your head around. The [docs page](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_scopes) helps a bit. But ultimately the `$script` prefix ties that variable to the script scope which inside a module is effectively module scoped for as long as that module is loaded in memory.

In any case, this looks good. Thanks again!

Issue #57, opened 2018-07-06
Azure Plugin incorrectly evaluates token expiration time


Cavorter:
When using the Azure plugin the expiration for the authentication token is evaluated incorrectly by simply calling `Get-Date` and comparing the localized TimeDate object against the UTC value attached to the token rather than converting the resulting TimeDate object to it's UTC value.

https://github.com/rmbolger/Posh-ACME/blob/55abb2d18832729e7dbebf61ce623fc97f88b9fe/Posh-ACME/DnsPlugins/Azure.ps1#L314

Cavorter:
FYI - I've got a fork up with the fix just taking a look to see if there are any tests around this I can poke at while I'm in here. :-)

rmbolger:
Nice! Thanks for the catch and fix. I'll follow up in the PR.

PR #58, opened 2018-07-06
Fixed Connect-AZTenant to evaluate token expiration correctly

Cavorter:
Fix for Issue #57 and added a few tests around the issue for good measure.

### DIFF
--- a/Posh-ACME/DnsPlugins/Azure.ps1
+++ b/Posh-ACME/DnsPlugins/Azure.ps1
@@ -286,9 +286,9 @@ function ConvertFrom-AccessToken {
     }
 
     # make sure the token hasn't expired
-    $expiresDT = ( Get-Epoch ).AddSeconds($claims.exp)
-    if ($expiresDT -lt [DateTime]::UtcNow) {
-        throw "The provided access token has expired as of $("{0:yyyy-MM-ddTHH:mm:ssZ}" -f $expiresDT)"
+    $expires = [DateTimeOffset]::FromUnixTimeSeconds($claims.exp)
+    if ((Get-DateTimeOffsetNow) -gt $expires) {
+        throw "The provided access token has expired as of $($expires.ToString('u'))"
     }
 
     # return an object that contains the 'expires_on' property along with the token
@@ -317,7 +317,7 @@ function Connect-AZTenant {
     # just return if we already have a valid Bearer token
     if ($script:AZToken ) {
         Write-Debug "Token Expires: $($script:AZToken.Expires)"
-        if ( ( Get-UtcDate ) -lt $script:AZToken.Expires) {
+        if ((Get-DateTimeOffsetNow) -lt $script:AZToken.Expires) {
             Write-Debug "Existing token has not expired."
             return
         }
@@ -360,7 +360,7 @@ function Connect-AZTenant {
 
     # create a token object that we can use for subsequence calls with a 5 min buffer on the expiration
     $script:AZToken = [pscustomobject]@{
-        Expires    = ( Get-Epoch ).AddSeconds($token.expires_on - 300)
+        Expires    = [DateTimeOffset]::FromUnixTimeSeconds($token.expires_on).AddMinutes(-5)
         AuthHeader = @{ Authorization = "Bearer $($token.access_token)" }
     }
 }
@@ -449,19 +449,4 @@ function Get-AZTxtRecord {
         $rec = @{id = $recID; name = $relName; properties = @{fqdn = "$RecordName."; TXTRecords = @()}}
         return $rec
     }
-}
-
-
-function Get-UtcDate {
-    [CmdletBinding()]
-    [OutputType([DateTime])]
-    Param( [DateTime]$Date )
-    Write-Output ( Get-Date @PSBoundParameters ).ToUniversalTime()
-}
-
-function Get-Epoch {
-    [CmdletBinding()]
-    [OutputType([DateTime])]
-    Param()
-    Write-Output ( Get-Date '1/1/1970' )
 }
--- a/Posh-ACME/DnsPlugins/Azure.ps1
+++ b/Posh-ACME/DnsPlugins/Azure.ps1
@@ -463,5 +463,5 @@ function Get-Epoch {
     [CmdletBinding()]
     [OutputType([DateTime])]
     Param()
-    Write-Output ( Get-UtcDate -Date '1/1/1970 00:00:00 GMT' )
+    Write-Output ( Get-Date '1/1/1970' )
 }
--- a/Posh-ACME/DnsPlugins/Azure.ps1
+++ b/Posh-ACME/DnsPlugins/Azure.ps1
@@ -40,7 +40,7 @@ function Add-DnsTxtAzure {
     }
 
     # build the record update json
-    $recBody = @{properties=@{TTL=10; TXTRecords=$txtVals}} | ConvertTo-Json -Compress -Depth 5
+    $recBody = @{properties=@{TTL=10;TXTRecords=$txtVals}} | ConvertTo-Json -Compress -Depth 5
 
     Write-Verbose "Sending updated $($rec.name)"
     Write-Debug $recBody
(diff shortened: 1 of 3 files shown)


rmbolger:
Yay, tests! You are awesome!

So in reviewing the rest of the code expiration related code, should we be doing more explicit UTC conversions on creation as well? Like here:

https://github.com/rmbolger/Posh-ACME/blob/55abb2d18832729e7dbebf61ce623fc97f88b9fe/Posh-ACME/DnsPlugins/Azure.ps1#L349-L353

`(Get-Date '1/1/1970')` returns a datetime with "Kind = Unspecified" but which seems to be neither local or UTC when comparing it to `.ToLocalTime()` and `.ToUniversalTime()` versions. We probably need to change that to something like `(Get-Date '1970-01-01T00:00:00Z').ToUniversalTime()` to make sure we're adding seconds to the right value? Or maybe the `ToUniversalTime()` isn't needed because comparisons don't count time zones?

Time is confusing.

rmbolger:
I went a little crazy and fleshed out the tests to cover more possible scenarios and verify the $script:AZToken object that gets created. I *think* changing the fix to where the `Expires` value is created is more correct than working around the issue where the check happens. I'm certainly open to being wrong though.

What do you think?

Cavorter:
Good catch on how AZToken is getting generated but I'm not sure that's entirely correct just yet. Just as an experiment I tried the following:
```
PS C:\src\Posh-ACME> Get-Date '1970-01-01T00:00:00Z' | Select-Object *

DisplayHint : DateTime
DateTime    : Wednesday, December 31, 1969 6:00:00 PM
Date        : 12/31/1969 12:00:00 AM
Day         : 31
DayOfWeek   : Wednesday
DayOfYear   : 365
Hour        : 18
Kind        : Local
Millisecond : 0
Minute      : 0
Month       : 12
Second      : 0
Ticks       : 621355752000000000
TimeOfDay   : 18:00:00
Year        : 1969

PS C:\src\Posh-ACME> ( Get-Date '1970-01-01T00:00:00Z' ).ToUniversalTime() | Select-Object *

DateTime    : Thursday, January 1, 1970 12:00:00 AM
Date        : 1/1/1970 12:00:00 AM
Day         : 1
DayOfWeek   : Thursday
DayOfYear   : 1
Hour        : 0
Kind        : Utc
Millisecond : 0
Minute      : 0
Month       : 1
Second      : 0
Ticks       : 621355968000000000
TimeOfDay   : 00:00:00
Year        : 1970

PS C:\src\Posh-ACME> ( Get-Date '1970-01-01T00:00:00Z' ).ToUniversalTime().Ticks -eq ( Get-Date '1970-01-01T00:00:00Z' ).Ticks
False
```

There's a few really odd things to me there. First that the object that is not derived from the `.ToUniversalTime()` method has a date of 12/31/1969 12:00:00 AM and that the ticks of the two objects are not equal. That says to me that something weird is going on if we aren't explicitly doing the conversion to UTC.

I also have an idea about streamlining the tests a touch instead of having so much repeated code. Should have a commit pretty soon.

Cavorter:
Nevermind, your original assertion that time is weird has been borne out more than I would like. The more I messed with this the worse it got and your changes appear to be working fine without the explicit UTC conversion so that seems like the best bet at this point. :-)

rmbolger:
What's funny is that I'm pretty sure I did that exact test as well and was equally confused. I think the key is that adding a time to the string that you pass to `Get-Date` always gives you a Local time regardless of the time zone that might be in the string.

```
> # from my local system currently on PDT (GMT-7)
> (Get-Date '2018-07-04T20:00:00Z'),(Get-Date '2018-07-04T13:00:00-07:00') | select Kind,DateTime

 Kind DateTime
 ---- --------
Local Wednesday, July 4, 2018 1:00:00 PM
Local Wednesday, July 4, 2018 1:00:00 PM

> (Get-Date '2018-07-04T20:00:00Z') -eq (Get-Date '2018-07-04T13:00:00-07:00')
True
```

What seems to be important when comparing the objects is that they have the same value for `Kind` whether they're both Local or both Utc (and not Unspecified). So we just need to make sure all of the DateTime instances are consistent.

Here's the other thing I noticed. It seems like trying to convert an Unix/Epoch time to a DateTime should only be done against a UTC time. I still can't really wrap my head around *why* this is. But the tests speak for themselves.

```
> # Correct: convert UTC then Add seconds
> (Get-Date '1970-01-01T00:00:00Z').ToUniversalTime().AddSeconds(1530734400) | select Kind,DateTime

Kind DateTime
---- --------
 Utc Wednesday, July 4, 2018 8:00:00 PM

> # Incorrect: Add seconds then convert UTC
> (Get-Date '1970-01-01T00:00:00Z').AddSeconds(1530734400).ToUniversalTime() | select Kind,DateTime

Kind DateTime
---- --------
 Utc Wednesday, July 4, 2018 7:00:00 PM
```

So if we want to work with Local times, we need to convert to UTC, add seconds, and then convert back to Local (Or just standardize on UTC and make sure everything compared has been converted first). And I think that means we still have some fixes to make. But I'd love to make a test that highlights this particular problem though.

Cavorter:
Seems pretty clear that converting to UTC first is going to be the right call in all cases. Taking a look through and there's a place in ConvertFrom-AccessToken where seconds are being added and then compared against a UTC type object so that'll have to get fixed too. So with multiple calls to Get-Date and DateTime objects seems like time to break it out into a function for simplicity.

Cavorter:
Hopefully this resolves it all, let me know if there's anything obvious missing.

Cavorter:
So I was talking with Tim Curwick at the local PowerShell User Group meeting last night and mentioned our issue to him and he swears that we should be using `Get-Date '1/1/1970'` without messing with `.ToUniversalTime()` and it should work reliably. Apparently it's either a magic string or the lack of a time zone indicator means that it gets treated as UTC unless you do something to make it think it's not. I'm going to go ahead and revert back to using that and hopefully we should be ready to go.

rmbolger:
I don't think that's right. If anything, the docs indicate `Unspecified` is treated like Local time rather than Utc which then suffers from being wrong when adding the epoch seconds directly to it like the previous example.

I was reading a bit last night about the difference between `DateTime` and [`DateTimeOffset`](https://msdn.microsoft.com/en-us/library/system.datetimeoffset(v=vs.110).aspx) [here](https://stackoverflow.com/questions/264617/when-would-you-prefer-datetime-over-datetimeoffset) and [here](https://stackoverflow.com/questions/4331189/datetime-vs-datetimeoffset). There seems to be a consensus that `DateTimeOffset` is the better fit in most cases, particularly this one where you're trying to compare absolute instants in time. Additionally, it has built-in methods for converting to/from Unix time.

I'm gonna play around with this a bit more.

rmbolger:
So the biggest benefit in this case for using `DateTimeOffset` over `DateTime` is that Local vs UTC becomes a non-issue during comparisons. 

```
> [DateTimeOffset]::Now -eq [DateTimeOffset]::UtcNow
True

> $localDTO = [DateTimeOffset]::Parse('2018-07-04T13:00:00-07:00')
> $utcDTO = [DateTimeOffset]::Parse('2018-07-04T20:00:00Z')
> $localDTO -eq $utcDTO
True
```

The built-in static `FromUnixTimeSeconds` is also super handy. The only downside is that you can't mock .NET methods directly with Pester. So I wrapped `[DateTimeOffset]::Now` in a private `Get-DateTimeOffsetNow` function. There will undoubtedly be other plugins that need this in the future. So it makes sense to just add it to the global list of private functions.

rmbolger:
I'm inclined to call this done at this point. Any objections?

I'll probably go back through the rest of the module at some point and evaluate moving to `DateTimeOffset` in the rest of the places that make sense. But that doesn't need to be part of this PR.

rmbolger:
Apologies for not waiting. Found a relatively critical bug in 2.4 this evening and rather than pushing a 2.4.1 hotfix just for it, I wanted to roll it into the 2.5 changes (which this is a part of) and push that instead.

Issue #53, opened 2018-06-23
GoDaddy Plugin fails removing acme challenge


davehope:
The GoDaddy plugin fails to remove DNS records from my zone (hope.mx), failing as follows:

```
Invoke-RestMethod : {"code":"INACTIVE_DOMAIN","message":"The given domain is not eligible to have its nameservers changed"}
At C:\Program Files\WindowsPowerShell\Modules\Posh-ACME\2.4.0\DnsPlugins\GoDaddy.ps1:125 char:17
+     $response = Invoke-RestMethod -Uri "$apiRoot/$zone/records" `
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (System.Net.HttpWebRequest:HttpWebRequest) [Invoke-RestMethod], WebException
    + FullyQualifiedErrorId : WebCmdletWebResponseException,Microsoft.PowerShell.Commands.InvokeRestMethodCommand
```

This looks to be because the GoDaddy API doesn't support the removal of individual records. The plugin currently implements this by overwriting the entire zone.

My untested assumption is that this fails for my domain since it is locked with GoDaddy, meaning the NS records cannot be updated.

davehope:
Since the GoDaddy API doesn't support the removal of individual records, the options as I see it are as follows:

1.  Continue with the current logic of replacing the entire zone (minus the record to be deleted) and update the readme. Possibly catch the exception and display something more friendly explaining why things failed. Those people with a locked domain will need to unlock their domains.

2. Investigate the potential of using the GoDaddy Domains API to unlock the domain, make the change and then re-lock the domain. This is risky as the subsequent lock call could fail, leaving the domain in a less-secure state. 

3. Replace the ACME challenge value with $null or similar

I'm happy to implement these, or another option, but would like to get a consensus on the preferred approach. My preference is the third option.

rmbolger:
Man, that is a lame limitation on GoDaddy's part. I agree about option 3. It seems to be the least risky and will ultimately allow more people to use the plugin without hassle.

@Dule11 and @Rukas are the other users of the GoDaddy plugin I know of. You guys have an opinion?

[...]

PR #71, opened 2018-08-15
 Change GoDaddy DNS Provider to not overwrite zone

davehope:
This PR makes changes to the behavior of the GoDaddy DNS Plugin. Specifically Remove-DnsTxtGoDaddy()

Prior to this PR, the functionality was:
- Download the list of records in the zone;
- Iterate over them looking for the relevant acme-challenge and remove it;
- Replace the the entire zone contents with the new zone;

For at least me, this results in an error ( see issue #53 ). Unfortunately, the GoDaddy Domains API does not support either the removal of an individual record, or changing the value of an existing record. The new functionality is therefore:

- Download the list of records in the zone;
- Iterate over them to see if there's a corresponding acme-challenge record;
- Replace all DNS Records for the specified Domain with the specified Type and Name with a single space;

Following these changes, the specific issue I was encountering no longer occurs.


rmbolger:
I made some changes that should hopefully make the code a bit more clear and consistent between the Add/Remove functions. In all cases now, we're only ever querying and acting on TXT records for the FQDN in question. I also added some caching in `Find-GDZone` so it doesn't have to do multiple lookups for the same zone in the same session.

Give this a run and see if it's still working for you. I tested against a random $0.99 domain I picked up today. But it has basically nothing in it except the defaults.

rmbolger:
The basic logic for both Add/Remove now looks like this:

- Get only the TXT records in the zone that match `$RecordName`
- Check if `$TxtValue` matches one of the data values in the set
- On Add, if record needs adding, append a new record to the existing set and send the new set.
- On Remove
  - If record needs removing and it's the last one left, send an empty data value
  - If record needs removing and it's not the last one left, remove the record from the existing set and send what's left

rmbolger:
Sweet. Thanks so much for your time and effort on this, @davehope. I'm sure it will benefit others as well.

PR #78, opened 2018-09-04
Add support for DeSEC DNS Hosting

nazar554:
https://desec.io/#!/en/product/dnshosting
https://desec.io/docs.html

TODO: support batch API

### DIFF
--- a/Posh-ACME/DnsPlugins/DeSEC-Readme.md
+++ b/Posh-ACME/DnsPlugins/DeSEC-Readme.md
@@ -12,5 +12,9 @@ The API token will be used with the `DSToken` parameter. We just have to create 
 `DSTTL` is the TTL of new `TXT` record (optional, defaults to 300 if not provided).
 
 ```powershell
-New-PACertificate test.example.com -DnsPlugin DeSEC -PluginArgs @{DSToken=(Get-Credential); DSTTL=3600}
+# if on Windows
+$token = Read-Host "API token" -AsSecureString
+New-PACertificate test.example.com -DnsPlugin DeSEC -PluginArgs @{DSToken=$token; DSTTL=3600}
+# otherwise
+New-PACertificate test.example.com -DnsPlugin DeSEC -PluginArgs @{DSTokenInsecure='yourdesectoken'; DSTTL=3600}
 ```
--- a/Posh-ACME/DnsPlugins/DeSEC-Readme.md
+++ b/Posh-ACME/DnsPlugins/DeSEC-Readme.md
@@ -9,9 +9,8 @@ No setup required, just use the the API token from registration email.
 ## Using the Plugin
 
 The API token will be used with the `DSToken` parameter. We just have to create it as a SecureString first.
-`DSTTL` is the TTL of new `TXT` record.
+`DSTTL` is the TTL of new `TXT` record (optional, defaults to 300 if not provided).
 
 ```powershell
-$dsToken = Read-Host "API token" -AsSecureString
-New-PACertificate test.example.com -DnsPlugin DeSEC -PluginArgs @{DSToken=$dsToken; DSTTL=3600}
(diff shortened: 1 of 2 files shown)


rmbolger:
Thanks, @nazar554! This looks great. I'll try to run some tests soon and get it merged. Or did you want me to wait for the batch API stuff?

rmbolger:
Or perhaps I won't be running tests quite yet. It seems the DNS hosting is still in beta. I signed up for beta access, but not sure when they'll get back to me.

rmbolger:
Got my beta API key and I can test now. 

First question, can we remove the `DSTTL` required parameter and just hard code a default value for everything? I don't think Let's Encrypt actually adheres to TTLs for the purposes of validation and it would probably confuse a lot of end users.

Second, I think I ran into a bug. When creating a wildcard certificate that contains both the wildcard name, `*.example.com`, and the root name, `example.com`, the Add function seems to throw an error creating the second TXT record.

```
Invoke-RestMethod : {"__all__":["R rset with this Domain, Subname and Type already exists."]}
At C:\Users\user1\Documents\WindowsPowerShell\Modules\Posh-ACME\DnsPlugins\DeSEC.ps1:42 char:13
+             Invoke-RestMethod "https://desec.io/api/v1/domains/$($dom ...
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (System.Net.HttpWebRequest:HttpWebRequest) [Invoke-RestMethod], WebExc
   eption
    + FullyQualifiedErrorId : WebCmdletWebResponseException,Microsoft.PowerShell.Commands.InvokeRestMethodCommand
```

nazar554:
Done, fixed second issue by escaping `...` with `%2e%2e%2e`. Also switched token to `[pscredential]` to make it work with .NET Core

rmbolger:
Unfortunately, `[pscredential]` has the same problem as `[securestring]` on non-Windows OSes. What I've been doing is creating an alternate parameter set for "Insecure" versions of normally secure parameters. For instance, here's the param definition for the Aliyun plugin I just implemented:

```powershell
function Add-DnsTxtAliyun {
    [CmdletBinding(DefaultParameterSetName='Secure')]
    param(
        [Parameter(Mandatory,Position=0)]
        [string]$RecordName,
        [Parameter(Mandatory,Position=1)]
        [string]$TxtValue,
        [Parameter(Mandatory,Position=2)]
        [string]$AliKeyId,
        [Parameter(ParameterSetName='Secure',Mandatory,Position=3)]
        [securestring]$AliSecret,
        [Parameter(ParameterSetName='Insecure',Mandatory,Position=3)]
        [string]$AliSecretInsecure,
        [Parameter(ValueFromRemainingArguments)]
        $ExtraParams
    )
```

So you can either specify `-AliSecret` as a secure string or `-AliSecretInsecure` as a normal string. The `DefaultParameterSetName` is also important. If they forget to include either, it prompts for the one we'd prefer which is the secure one. And if they include both, it will also use the secure one. 

I'd suggest making changing `-DSToken` back to a securestring and adding a `-DSTokenInsecure` like the above example. Then since you end up using a normal string for the token anyway, the first thing you can do inside the function is something like this and then always use `$DSTokenInsecure` from that point forward.

```
# convert the secure token to a normal string
if ('Secure' -eq $PSCmdlet.ParameterSetName) {
    $DSTokenInsecure = (New-Object PSCredential ("user", $DSToken)).GetNetworkCredential().Password
}
```

rmbolger:
Also, good catch on the `...` escaping.

nazar554:
Done, both parameters work when using `pwsh.exe` on Windows

rmbolger:
Awesome. Everything working on my end. I added a couple tiny comment-help fixes. This should go live in the next release.

PR #94, opened 2018-10-19
Fix DNS clean-up when TXT records are not created for all domains

philr:
`$toValidate` is an array of domain indexes. The [DNS clean-up code](https://github.com/rmbolger/Posh-ACME/blob/655b15e97f8878f5b236477553bf356653e6796b/Posh-ACME/Public/Submit-ChallengeValidation.ps1#L160) is currently iterating over `$toValidate`, but using the loop variable to identify domains instead of looking up the domain index in `$toValidate`.

If a certificate was requested for two domains, but the first was already validated, a DNS record would only have been created for the second domain. The clean-up code would then try and delete a non-existent DNS record for the first domain instead of the record that was created for the second domain.

This pull request changes the DNS clean-up code to use the index from `$toValidate` to access the `$allAuths`, `$DnsAlias` and `$DnsPlugin` arrays.

### DIFF
--- a/Posh-ACME/Public/Submit-ChallengeValidation.ps1
+++ b/Posh-ACME/Public/Submit-ChallengeValidation.ps1
@@ -159,12 +159,14 @@ function Submit-ChallengeValidation {
     } finally {
         # always cleanup the TXT records if they were added
         for ($i=0; $i -lt $toValidate.Count; $i++) {
-            if ([string]::IsNullOrWhiteSpace($DnsAlias[$i])) {
+            $index = $toValidate[$i]
+
+            if ([string]::IsNullOrWhiteSpace($DnsAlias[$index])) {
                 # unpublish normally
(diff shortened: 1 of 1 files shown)


rmbolger:
Oh man, this is a great find! I totally screwed that up. Thanks again, @philr.

rmbolger:
In retrospect, it would probably be easier to understand iterating directly over the `$toValidate` array rather than an indexed for loop. Something like this:

```
foreach ($index in $toValidate) {
    // unpublish stuff
}
```

I was clearly not thinking clearly when I originally wrote it. This is fine for now though.

philr:
Yes, using `foreach` would have been a bit simpler. I don't spend enough time in PowerShell, so missed that one.

Thank you for all your work on Posh-ACME. I've been using it to automate the issuing of certificates on deployment of an Azure Cloud Service. Apart from these minor issues I raised, it has been working perfectly.

PR #96, opened 2018-10-19
URL-encode the Azure client id and secret

philr:
The content type of the body of the `oauth2/token` request made in the Azure DNS plugin is `application/x-www-form-urlencoded`. The `client_id` and `client_secret` parameters must therefore be URL-encoded to allow characters such as '&' to be used in values.

This pull request adds the necessary url-encoding.


rmbolger:
Love it. Thanks, @philr.

PR #98, opened 2018-10-24
Allow creation of certificates from CSR data

virot:
Okey probably should have done issue, but hey, this was more fun.

I have been using it to create letsencrypt certificates for my HP servers. The HP ILO does not allow uploading of keys. So I needed to use the generated CSRs.

This was a quick hack to allow New-PACertificate to use a CSR instead of creating it from scratch.

### DIFF
--- a/Posh-ACME/Public/New-PACertificate.ps1
+++ b/Posh-ACME/Public/New-PACertificate.ps1
@@ -67,8 +67,7 @@ function New-PACertificate {
     # - is pending, but expired
     # - has different KeyLength
     # - has different SANs
-    $order = $null
-    try { $order = Get-PAOrder $Domain[0] -Refresh } catch {}
+    $order = Get-PAOrder $Domain[0] -Refresh
     $SANs = @($Domain | Where-Object { $_ -ne $Domain[0] }) | Sort-Object
     if ($Force -or !$order -or
         $order.status -eq 'invalid' -or


rmbolger:
Hey @virot. Thanks for this! It seems like a great start for this feature. I have some questions/comments.

It looks like you have the `-CSR` param of `New-PACertificate` expecting the string contents of a CSR file. My gut says it would be easier for users to specify a file path and let the module read/parse it and store it on the order object. I'd probably also move that parameter and logic into `New-PAOrder` so that people doing things manually can still provide their own CSR. You'd still have the param in `New-PACertificate`, but it would just pass it as-is to `New-PAOrder`.

The changes in `Export-PACertFiles` make it dependent on having an order object which it wasn't previously. Instead, you could either add a switch like `[switch]$NoKeyExports` or just check if the `$keyFile` file exists...or maybe both.

How do renewals work? Can you renew with the same CSR over and over? Or do you need to generate a new one each time? Like, does Submit-Renewal need to skip orders that have a specified CSR because they'll fail?

virot:
Its me again..
So I did a few code revisions. but I have no clue on how to update this pull request.. But if you check this page you can see the new versions I did.

https://github.com/rmbolger/Posh-ACME/compare/master...virot:CSR

Also a CSR can be used as many times as one wants.. BUT it depends on the key user if you can use the key multiple times, I know lots of loadbalancers allow it.

rmbolger:
Normally to update a pull request, you'd just make new commits to the branch that you created the pull request from originally. But it looks like you may have deleted that original branch? No big deal though. You can just submit a new one against the new branch and I'll close this one and link it to the new one.

virot:
There started to be to much junk since I didnt really know how to use branches on github :)

rmbolger:
This PR was replaced by PR #99

PR #99, opened 2018-10-25
Allow generating certificates from pre-made CSR

virot:
So you can create certificates from CSRs


rmbolger:
See PR #98 for previous context.

rmbolger:
This is going to take me a bit to review since it potentially affects large portions of core functionality. I truly appreciate the effort and want to make sure we get it right the first time.

virot:
Really sorry for the ugly part of Extract-DomainsFromCSR.
But I had no clue how to go down the rabbit hole to the dnsnames. First time with BouncyCastle. Or well.. Its Powershell but I couldnt figure out how the data was wrapped.

rmbolger:
Not a problem. It’s not the easiest thing to deal with with considering the lack of docs for it. But I’ve already got some tweaks to clean it up in my local copy.

rmbolger:
I refactored `Extract-DomainsFromCSR` into `Get-CsrDetails` and made a few other changes. I haven't had a chance to test things much yet and the old Pester tests for Extract-DomainsFromCSR still need updating. But I probably won't be able to work on this much more for the next few days.

virot:
I have no clue how to get your refactoring into my code :) but I read through everything.
I hope atleast I helped a little :)

rmbolger:
If you haven't made any local changes since your "Fixed stupid bug." commit, you should just be able to `git pull` and bring down my changes into your local copy. If you have, it's a bit more difficult depending on what changes you made. But I can walk you through it if you want.  You could also just delete your local branch and re-checkout the upstream copy like this:

```
git checkout master
git branch -D CSR
git checkout CSR
```

In any case, yes. You definitely helped! I'm sure there are others who will appreciate this feature when it's done.

virot:
Trying to understand rfc5280, Is CN in Subject a MUST. In that case failing test should throw. Otherwise we need to figure out why it says 4 out of 3.

rmbolger:
According to [this](https://security.stackexchange.com/questions/55414/is-the-common-name-mandatory-for-digital-certificates), it’s not required but definitely rare for a cert/req to not have one. 

We may have to modify the header check again for CSRs in `Import-Pem`. The requests I’m generating from Windows do include “NEW” so it seems there’s some variability there.

Issue #101, opened 2018-10-31
Request for adding IBM Softlayer DNS Provider


chandan1001:
Softlayer has python API for managing DNS 

https://softlayer-python.readthedocs.io/en/latest/api/managers/dns.html

rmbolger:
It looks like Softlayer is now owned by IBM under their IBM Cloud umbrella. Is that correct? Would a plugin more appropriately be named something IBM related now?
http://www.softlayer.com/

As for a plugin. It wouldn't make sense to use a Python API from PowerShell. But it looks like there's a REST API I could call directly.
https://console.bluemix.net/docs/infrastructure/dns/dns-api.html#getting-started-with-the-dns-api-
https://softlayer.github.io/article/rest/

I'll see what I can do after I finish up with some other issues.

chandan1001:
@rmbolger yes softlayer is owned by IBM now. They give multiple names Bluemix, IBM Cloud etc. but we still call it  Softlayer. You can name it IBM Softlayer.

rmbolger:
I just pushed a branch with a first stab at a SoftLayer plugin if you wanna give it a try. It seems to be working for me. But I'd love a second opinion.

https://github.com/rmbolger/Posh-ACME/blob/ibmsoftlayer/Posh-ACME/DnsPlugins/IBMSoftLayer.ps1
https://github.com/rmbolger/Posh-ACME/blob/ibmsoftlayer/Posh-ACME/DnsPlugins/IBMSoftLayer-Readme.md

rmbolger:
I decided to just merge it so it should be easier to test now. But please do let me know if it works for you. I'll leave the issue open just in case.

chandan1001:
DNS plugin seems to be working. Can you reduce the TTL from 900 to 60 ? I had to make this change in the API. 
Thanks

PR #104, opened 2018-11-03
Add IBMSoftLayer plugin and readme

rmbolger:
This change adds a plugin for [IBM Cloud DNS](https://www.ibm.com/cloud/dns) (formerly known as SoftLayer).

It includes two parameter sets that distinguish between `-IBMCredential` for Windows and `-IBMUser`/`-IBMKey` for non-Windows so folks can still use it in a "less secure" mode.


rmbolger:
This should close issue #101

PR #120, opened 2019-01-18
Add name.com DNS plugin

ravensorb:
Implemented a new DNS plugin for name.com


rmbolger:
Hey @ravensorb, thanks for this! I've got a few style related tweaks mostly around debug/verbose logging that I'll commit shortly. Then I'll do a couple tests. It looks like Name.com has a testing environment available. I may add a parameter to optionally use that in the plugin to make testing easier.

ravensorb:
Not at all -- I thought I had caught all of those test cases so thanks!

PR #142, opened 2019-05-13
Ignore Get-Content errors in Import-PAConfig rather than SilentlyIgnore

maybe-hello-world:
I suggest changing EA:SilentlyContinue to EA:Ignore in this file because it looks like not very important if this command would not be executed successfully but SilentlyContinue append error message to $Error and after it's rather hard to automatically parse $Error and decide whether script failed with an important reason or it can be ignored.
Didn't checked other files, sorry :( (I think the similar problem could be somewhere else)


rmbolger:
You are correct that the `Get-Content` calls failing in this function should generally be ignored. I didn't even think about how the errors would pollute $Error. Technically, someone might care if it's an error other than that the file doesn't exist. But that's a pretty significant edge case and I'm ok with this change.

PR #145, opened 2019-05-23
Fix case sensitivity issue when record name case doesn't match zone case on the server

Makr91:
Also adding optional parameter to specify the DNS Zone

### DIFF
--- a/Posh-ACME/DnsPlugins/Windows.ps1
+++ b/Posh-ACME/DnsPlugins/Windows.ps1
@@ -13,6 +13,7 @@ function Add-DnsTxtWindows {
         [pscredential]$WinCred,
         [switch]$WinUseSSL,
         [string]$WinZoneScope,
+        [string]$WinZone,
         [Parameter(ValueFromRemainingArguments)]
         $ExtraParams
     )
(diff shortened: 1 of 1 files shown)


rmbolger:
Hey @Makr91, that's for the PR on the case sensitivity issue. Usually text comparisons are case-insensitive by default in PowerShell, so this surprised me. Give me a bit to do some testing myself and I'll get this in.

However, I think the WinZone optional parameter should be removed because the parameters for a plugin get saved for the account rather than the specific certificate being requested. So if you create a new cert in a different domain, that now saved WinZone parameter would be applied to the new cert as well and be wrong. I appreciate the intention though.

rmbolger:
Ah ha, I get it now. So it wasn't the comparisons that were case-sensitive, but the `.Replace` method splitting the zone name from the record portion. I'm definitely going to have to do some searching for other instances where this could potentially be an issue.

I hope you don't mind. I fixed the issue in a slightly different way and removed the WinZone parameter. Thanks again for the heads up on this though.

Makr91:
No Problem, I wanted to try to fix an issue where I could, without just complaining!

rmbolger:
It's all good. For future reference though, submitting an issue is definitely not complaining unless the text of the issue is literally whining with no useful info about the problem. Some projects may even require an issue exists to reference a pull request against. In any case, thanks again.

PR #149, opened 2019-06-17
Adding Simple DNS Plus Plugin for DNS Validation

alphaz18:
Hi, I think I have done this right, 
This is just to add SimpleDNSPlus Plugin. Simple DNS Plus is a 
DNS Server Software vendor that is used by some people and we happen to use it.


rmbolger:
Hey @alphaz18. Thanks for this! There are a few things I noticed just quickly skimming the PR that could use fixing. It looks like the commit caught a couple files unnecessary files in the `.vs` folder. Removing those would be great. The readme also has a reference to generating a GoDaddy API key in the Setup section that probably doesn't need to be there.

It looks like SimpleDNS has a trial version, so I may download it and do some quick testing of the plugin before I merge the PR.

alphaz18:
I tested bits and pieces of the PS1 Commands, because I wasn't sure how to test the whole thing as I am not sure how to install it from my fork, so I hope it works! :D . but yes they do have trial and feel free to test it

rmbolger:
For future reference, there are two common ways to test your fork. You can just copy the plugin file into the DnsPlugins folder of your installed copy of the module and then do another `Import-Module Posh-Acme -Force`.  You can also just run the `instdev.ps1` script from your forked copy to have it install your whole forked version of the module into your personal profile and takes care of the re-import automatically.

Then if you want to test just the plugin functionality without needing to generate an actual certificate, you can just use `Publish-DnsChallenge` and `Unpublish-DnsChallenge` (and `Save-DnsChallenge` if used)   like this using whatever `$pArgs` you have setup:

```powershell
Publish-DnsChallenge example.com (Get-PAAccount) fakedata SimpleDNSPlus $pArgs
Save-DnsChallenge SimpleDNSPlus $pArgs
Unpublish-DnsChallenge example.com (Get-PAAccount) fakedata SimpleDNSPlus $pArgs
Save-DnsChallenge SimpleDNSPlus $pArgs
```

I tend to do 3x Publish with the same domain and different fake token strings and then 3x Unpublish to make sure I cover the different potential states the TXT record might be in on the server.

rmbolger:
I made some tweaks to bring the authentication parameters more in line with how other plugins work. `SdnsCred` is now in the default parameter set and takes a pre-made PSCredential object that will save encrypted to disk on Windows platforms. It's also not mandatory in case people have their API configured for anonymous authentication. `SdnsUser` and `SdnsPassword` are now in the "insecure" parameter set and will ultimately be used by non-Windows folks running PowerShell earlier than 6.2 or anyone who doesn't want their credentials encrypted on disk. 

I tweaked how you were dealing with self-signed certs to make use of some code that's already in the module and added an optional `SdnsIgnoreCert` option.

After playing with the server a bit, I realized that we also can't rely on any hard coded API root path. People could be using HTTP or HTTPS and any combination of sub-paths. So `SdnsServer` has been replaced with `SdnsApiRoot`.

There was a tiny bug I fixed in the check for existing TXT records (wasn't checking the value with embedded quotes).

I also fleshed out the readme a bit and added more examples for the different auth options.

It seems like a lot, but the basic flow of the API requests is the same. Thanks again for the contribution.

PR #153, opened 2019-06-21
Add Dreamhost DnsPlugin

jhendricks123:
Tested against my own Dreamhost account with success.

### DIFF
--- a/Posh-ACME/DnsPlugins/Dreamhost-Readme.md
+++ b/Posh-ACME/DnsPlugins/Dreamhost-Readme.md
@@ -4,7 +4,7 @@ This plugin works against the [Dreamhost](https://www.dreamhost.com/) API. It is
 
 ## Setup
 
-We need to generate the API key that will be used to update DNS records. Open the [Web Panel API](https://panel.dreamhost.com/index.cgi) page, and generate a new API key with dns-add_record and dns-remove_record permissions. It is recommended to name this API key something memorable like "Posh-ACME" by entering this in the "Comment for this key" field. Take note of your new API key.
+We need to generate the API key that will be used to update DNS records. Open the [Web Panel API](https://panel.dreamhost.com/index.cgi) page, and generate a new API key with dns-add_record, dns-list_records, and dns-remove_record permissions. It is recommended to name this API key something memorable like "Posh-ACME" by entering this in the "Comment for this key" field. Take note of your new API key.
 
 ## Using the Plugin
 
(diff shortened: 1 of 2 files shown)


rmbolger:
Hey @jhendricks123. Thanks for the plugin! Man, that DreamHost DNS api is rather...limited. I have just a few comments/requests before I merge this.

The security protocol tweaks you have in there shouldn’t be necessary. The module takes care of it conditionally on load. It’s conditional because the code isn’t compatible with PowerShell Core so it only needs to run in Desktop edition.

Can you add a basic usage guide similar to the others that exist? Include things like where to find the page to get your API key and example usage.

There are two common scenarios that most plugins handle that could use some additional error handling if I’m reading things correctly. Trying to add a record that already exists shouldn’t generate an error for the user. Trying to delete a record that doesn’t exist also shouldn’t generate an error. An easy way to test without needing to futz with actual cert stuff is just using `Publish-DNSChallenge` and `Unpublish-DNSChallenge` like this. 

```
$pArgs = @{DreamhostApiKey=‘xxxxxxxx’}
Publish-DNSChallenge example.com (Get-PAAccount) fake Dreamhost $pArgs
Publish-DNSChallenge example.com (Get-PAAccount) fake Dreamhost $pArgs
Unpublish-DNSChallenge example.com (Get-PAAccount) fake Dreamhost $pArgs
Unpublish-DNSChallenge example.com (Get-PAAccount) fake Dreamhost $pArgs
```

The first publish will add the record. The second one will fail from an api standpoint, but you should basically catch the “record exists” error and ignore it. The first unpublish will remove the record. The second will generate the “no record” error and you should just catch and ignore the same way.

jhendricks123:
I didn't realize it shouldn't be considered an error for a DNS record to already exist - I'll swallow those errors. Should I write-error on any other error encountered?

I'll write up a guide in the next few days following the others as a template and send the updates your way :)

rmbolger:
No worries. If we were making a more generic Dreamhost DNS API PowerShell wrapper, we probably would want to surface those errors. But for the purposes of a Posh-ACME plugin, we only care that the record we want to exist exists and the record we want to delete is gone. I'll usually catch those cases and just make a Write-Debug call that mentions the record already exists or was already deleted.

I should probably update the plugin dev guide to mention this as well.

rmbolger:
Tangent question. Do you know if Dreamhost supports hosting DNS without actually transferring your domain to them. Like if someone wanted to keep their existing registrar but still have Dreamhost do the DNS (and presumably web) hosting, do they support that?

jhendricks123:
Yep, all my domains are on GoDaddy and I only forward DNS over to Dreamhost.

jhendricks123:
Let me know what you think - I've removed the SSL/TLS tweaks, and now only throw on errors besides "record already exists" or "record not found".

At first I handled the record exists/doesn't exist errors by looking for a matching record in the dns-list_records response, but realized that was wasteful and just swallowed the record_already_exists* and no_such_* responses.

rmbolger:
Awesome. Thanks for the updates. Merging now.

PR #161, opened 2019-07-26
Add self hosted HTTP challenge server

Soltroy:
Should be functionally ready. Tested all different methods to call the function (without Parameter, with Parameter, Piped) within Linux docker containers.
Spelling of verbose messages and synopsis should be checked, English is not my native language.
Also I wasn't sure if I should increase versioning, left that as it is.


rmbolger:
This is really cool, @Soltroy. Thank you. I've only had a chance to skim the code so far. It's going to take a bit to play around with it.

Quick question. It looks like you're basically spinning up a separate listener for each authorization. Roughly how long does each loop take assuming you're not hitting TTL timeouts? Did you consider trying to spin up a single listener that would support answering all of the challenges simultaneously?

Soltroy:
thx @rmbolger

Absolutely correct, it spins up a listener for every authorization. It was easier that way to check validation. Within test environment (which responds immediately) listener always finished after 5 seconds (timespan for validation check). For clearence - Validation check is not done after every response but every 5 seconds from runtime. Reason for this is that I already got errors within testsystem because of too many requests to ACME server.

Soltroy:
@rmbolger hope you didn't do to much testing and code reading. Your mention of the separate listener didn't leave me any peace and I refactored the listener code. Now the listener only runs once which also results in a correct TTL. I'll do some code cleaning tomorrow and commit the changes.

rmbolger:
Hah, nice. Only skimmed enough to get the gist. I look forward to seeing the updated version. Hoping to get into it this week.

rmbolger:
Is the ability to use this with a SAN value from an existing order important to your workflow? I like the intent, but I'm worried it might run into weird edge cases where people have purposefully made multiple orders with the same set of SANs but different primary domains (usually to get different encryption types).

Also, is the current code supposed to be checking for SANs in any existing order or just the currently selected order? Because I think it's doing the latter right now.

Soltroy:
Absolutely not, it was just a 'what if someone passes a SAN' catching - never thought of the possibility of multiple "same SAN" orders.
I'd be absolutely fine to change the process to accepting only a MainDomain.
I think current code would return multiple MainDomains if SAN is present in multiple Orders - and would break the function.... - will add this to my testing scripts.
I see two options here,
1. Get rid of SAN checking and accept only MainDomain - more user responsibility but easier code
2. Check SAN against all orders and process all of them - more complex code and maybe http listener is fired up for a challenge which should be DNS validated (who knows what people are doing)
Personally I'd prefer option 1.

rmbolger:
Removing the SAN checking is what I was thinking for now. Simple is better. And if there are people who really need that, it's not terribly difficult to add the functionality in their own scripts.

I may also eventually add an option to allow for SAN checking directly in `Get-PAOrder`.

rmbolger:
So I started testing on my own stuff and things look really good so far. The biggest pain point right now is getting the HttpListener to not throw an exception when it starts due to prefix binding issues. And the issues you run into are different depending on whether you're on Windows or Linux.

The Linux implementation seems to be limited to only using [IPs that are actually on the box or names that resolve to one of those IPs](https://stackoverflow.com/questions/49267224/httplistener-the-requested-address-is-not-valid-in-this-context). So a Linux box with a private IP behind a NAT where the cert's FQDN resolves to a public address fails when using the default option to bind to the cert's FQDN. Wildcard binding with the `-NoPrefix` parameter also works.

Windows, on the other hand, doesn't seem to care what you use for the bind prefix as long as you do it from an elevated (admin) process. But if you don't, you need to pre-register (as admin) the exact prefix you're going to use with `netsh http add urlacl`.

I'm wondering if it makes sense to make wildcard prefix binding the default rather than the exception. On Windows, this would make adding the necessary urlacl easier for non-elevated processes because you can just add one instead of one for each name in the cert. And on Linux things would work regardless of whether the server has the direct internet-facing IP or not.

There's a warning in the [HttpListener doc page](https://docs.microsoft.com/en-us/dotnet/api/system.net.httplistener?view=netframework-4.8) that suggests avoiding wildcard binding due to "security vulnerabilities". But that seems to revolve around DNS rebinding attacks which aren't really a risk in this short lived listener.

With wildcard binding as the default, I'd probably add an optional parameter called something like `BindingPrefixes` that would accept a `[string[]]` of FQDNs/IPs. If defined, it would attempt to use those instead of the wildcard binding.

Also, regarding wildcard binding....do you understand the difference between `http://+/` and `http://*/`? Because I read the docs and I'm still pretty confused.

Soltroy:
Awesome, thanks for the feedback - I have to extend my testing....
`BindingPrefixes` with `string[[]]` and Wildcard per default sounds good, I will follow this path.

Windows binding and elevated rights - did some quick tests and now  I see two possibilities here,
1. check if function is called in windows environment, check if netsh is set (this hopefully works without cmd injection) and write error that elevated rights are mandatory / netsh has to be set
2. use localhost instead of wildcard - it seems like for localhost elevated rights are only needed if listener is running on port 80 - but this would require additional testing if the listener will answer to requests coming from outside

`http://+/` / `http://*/`
`*` doesn't answer if another process is using this port
`+` does answer, even if another process is using this port
Never tested this, but it reads like '+' would be an option for the force parameter. I'll do research here, especially if it really does override a present process....
Also I think I should do more extensive testing on MacOS, if Windows and Linux already behave that different, who knows what Mac is doing.

Next code update will take a while, I'm in my last week before vacation and have to finish a few things for work.

rmbolger:
No worries on the timing. Have a good vacation. I just got back from mine. I may push some changes with the stuff we've talked about because it's fun, not because I'm impatient.

rmbolger:
Also, it looks like localhost bindings on Windows do indeed only respond to queries from the localhost. So unfortunately, that option's out.

webprofusion-chrisc:
Hi, Certify The Web uses this same approach (temporary wildcard binding on the acme-challenge prefix) for it's built-in http listener. The most common problem to note is when users run something outside of http.sys on port 80 (so Apache, nginx etc) which prevents the listener from binding. Other than that it works extremely well.

rmbolger:
Hey @webprofusion-chrisc. Thanks for the heads up! Do you have a sense for the percentage of people running their Windows servers directly facing the internet (or behind NAT, but with direct port mapping) rather than behind a load balancer or reverse proxy? I've always been curious.

Ironically, I don't actually know that many people personally who run public websites from Windows. The few I run all sit behind a Linux HAProxy instance where I'm doing the SSL termination.

webprofusion-chrisc:
@rmbolger I don't have specific numbers but I confident  > 90% of windows servers that use Certify will be internet facing rather than having an appliance handling SSL. Most of users started without https and have had to introduce it since chrome etc started to require it more aggressively. Load balanced sites are in the minority however its becoming much more common to just put cloudflare in front of your site (which I do), which also removes the need to SSL directly on the server.

Regarding prefix, I'm currently using http://+:80/.well-known/acme-challenge as not using /.well-known/acme-challenge would eat all the http requests and potentially prevent them getting to IIS etc. 

Note there are now several other windows tools that do the same thing for acme challenges so occasionally you can also run into ones where the listener is 'always-on' and responding to challenges even if your client is the one being asked to do validation, which can be confusing for everyone. I suggest setting a Server header so you know when it's your app that's responding (and not someone else). Often when people are trying out acme clients they try a bunch of them either all together or over time.

rmbolger:
I suppose it makes sense that people who would be using a Windows based ACME client would be terminating SSL on their Windows servers. I'm dumb. Still curious about the general trend though outside of ACME.

I changed the function's wildcard default to include the `.well-known/acme-challenge` suffix and switch to the stronger `+` wildcard. If I understand the weaker `*` wildcard, I think it works sort of like a catch all where it only matches on things that haven't already been caught by "stronger" or more specific matches. In any case, this seems like a better default.

I'm not sure I understand about the server header. You're saying to add that in the response we send? Where would that be exposed when troubleshooting?

rmbolger:
@Soltroy I realize I went a little crazy with changes. My apologies. Let me know if anything doesn't work with how you intended to use this. I'm pretty excited about it though. For now in 3.x, it will be a nice to have feature for people doing HTTP challenges. But in 4.x, it will definitely form the basis for the first HTTP plugin once I figure out how best to architect those.

Soltroy:
@rmbolger No need for apologies - pretty interesting and amazing how this develops, I'm just a bit disappointed that I can't provide more input ATM.
But I do think that it's time to discuss the goal of the functionality. My initial intention was just a small http listener for people like me who don't operate a web server and just want to fire up ACME Http challenge (and don't want to operate a web server for just this task).
I think we should focus on two situations,
- web server is present
here, a plugin based challenge initiation would be IMHO the better way, no need to care about who fetches the requests
- no web server present
http listener. Urlacl and SSL would be nice, but if they get integrated I'd walk the hard way with direct interaction with API and avoid using netsh (I'm not a fan of embedded dos commands in PowerShell functions). But personally I think that this is something people can take care about themselves (of course with information from the cmdlet what they have to do)

In my mind forms a picture like
3.x - 'stupid' http listener which already uses a defined object (something like PoshACME.PAHttpChallenge) with all necessary information for http challenge. 
With version 4.x comes another cmdlet which uses the previously defined object as input and relays on web server plugins (IIS, Apache, .....) to interact with those.
Object creation can be done within the cmdlet or with a separate cmdlet (Get-PAHttpCHallenge).

Another way could be to get rid of Start-PAHttpCHallenge, change to a wrapper like Publish-PAHttpChallenge with just one plugin/integrated functionality (parameter integratedHttpListener) that does what Start-PAHttpChallenge is supposed to do. With Version 4. this cmdlet gets extended to use additional plugins just like the DnsPlugins. As long as command for firing up local listener doesn't change we even wouldn't have a function change with major increase.

rmbolger:
I think for 3.x, it's pretty close to done. I like the model of just using the current order unless otherwise specified. Those that are already doing HTTP challenges now have an additional tool to make that easier. Renewals aren't as easy as DNS with plugins and `Submit-Renewal`, but it's not terrible.

```powershell
# barebones cert creation and renewal workflow is the same like this
New-PAOrder example.com
Invoke-HttpChallengeListener
New-PACertificate example.com
```

4.x I had already planned on making a plugin model for HTTP challenges. Still not sure whether it'll be unified with the DNS plugins or separate. But the goal is that new certs and renewals become single commands just like DNS. This function would end up in a plugin called something like 'SelfHosted'. So you might use it something like this:

```powershell
New-PACertificate example.com -Plugin SelfHosted -PluginArgs @{Port=8080}
```

And then `Submit-Renewal` works just like it does with everything else. The other day 1 plugin I assume would exist is a simple file copy plugin. So you're already running an HTTP server via IIS, Apache, etc., and you just need the plugin to copy the challenge files to a specific location.

PR #165, opened 2019-08-10
Fix for Domeneshop when using insecure parameter

ornulfn:
Using insecure parameters will throw an error.
The PSCredential object is only needed to decrypt the secure string.

Tested using PS Core 6.2.0 on Linux

Signed-off-by: Ørnulf Nielsen <40027793+ornulfn@users.noreply.github.com>


rmbolger:
Good catch

PR #170, opened 2019-09-09
Support supplying new PluginArgs on renewal

matt-FFFFFF:
Consider scenarios where we are using short lived access tokens for authentication to a DNS provider. The previously supplied token will now no longer be valid, therefore it would be good to be able to supply a new set of PluginArgs in the renewal.

This scenario is applicable in CI/CD workflows. I have run into this issue using the Azure DNS plugin where I use az cli to generate an access token to supply to Posh-ACME. I did not want to supply standing credentials as they would be stored in the .config folder as I would have to manage that.

I realise this could be problematic when multiple DNS plugins are used. Another alternative is to make public the ```Export-PluginArgs``` function. This way people can update them before running the ```Submit-Renewal``` command.

### DIFF
--- a/Posh-ACME/Public/Submit-Renewal.ps1
+++ b/Posh-ACME/Public/Submit-Renewal.ps1
@@ -167,6 +167,9 @@ function Submit-Renewal {
     .PARAMETER NoSkipManualDns
         If specified, orders that utilize the Manual DNS plugin will not be skipped and user interaction may be required to complete the process. Otherwise, orders that utilize the Manual DNS plugin will be skipped.
 
+    .PARAMETER PluginArgs
+        A hashtable containing an updated set of plugin arguments to use with the renewal. So if a plugin has a -MyText string and -MyNumber integer parameter, you could specify them as @{MyText='text';MyNumber=1234}.
+
     .EXAMPLE
         Submit-Renewal
 
--- a/Posh-ACME/Public/Submit-Renewal.ps1
+++ b/Posh-ACME/Public/Submit-Renewal.ps1
@@ -9,7 +9,8 @@ function Submit-Renewal {
         [switch]$AllAccounts,
         [switch]$NewKey,
         [switch]$Force,
-        [switch]$NoSkipManualDns
+        [switch]$NoSkipManualDns,
+        [hashtable]$PluginArgs
     )
 
     Begin {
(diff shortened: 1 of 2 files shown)


Posh-ACME/Public/Submit-Renewal.ps1
```diff
@@ -9,7 +9,8 @@ function Submit-Renewal {
         [switch]$AllAccounts,
         [switch]$NewKey,
         [switch]$Force,
-        [switch]$NoSkipManualDns
+        [switch]$NoSkipManualDns,
+        [hashtable]$PluginArgs
```
rmbolger: Can you add an associated `.PARAMETER` help item down around line 170 with text similar to the one in `New-PACertificate`? Something like this:

> A hashtable containing an updated set of plugin arguments to use with the renewal. So if a plugin has a -MyText string and -MyNumber integer parameter, you could specify them as @{MyText='text';MyNumber=1234}.

Tutorial.md
```diff
@@ -162,6 +162,17 @@ Submit-Renewal -AllAccounts
 
 These are designed to be used in a daily scheduled task. **Make sure to have it run as the same user you're currently logged in as** because the module config is all stored in your local profile. Each day, it will check the existing certs for ones that have reached the renewal window and renew them. It will just ignore the ones that aren't ready yet.
 
+### Updating DNS Plugin Parameters on Renewal
+
+If you are authenticating to your DNS provider using a short-lived access token, you may need to update said token when it comes to renewal. You can do this by specifying the new plugin parameters using the ```-PluginArgs``` parameter. **The full set of plugin arguments must be specified**.
```
rmbolger: Thanks for including the update to the tutorial with this. I'd just tweak it slightly to include cases where normal credentials have changes as well.

```suggestion
DNS provider credentials can change over time and some plugins can be used with purposefully short-lived access tokens. In these cases, you can specify the new plugin parameters using the ```-PluginArgs``` parameter. **The full set of plugin arguments must be specified.**
```

rmbolger:
Hey @matt-FFFFFF. Thanks for the PR! It's so simple, yet so useful. I left a couple comments with some suggested tweaks to help/tutorial stuff. But other than that, it looks good.

Posh-ACME/Public/Submit-Renewal.ps1
matt-FFFFFF: Done - see 7805fee

matt-FFFFFF:
> Hey @matt-FFFFFF. Thanks for the PR! It's so simple, yet so useful. I left a couple comments with some suggested tweaks to help/tutorial stuff. But other than that, it looks good.

No worries, HTH.

matt-FFFFFF:
Hi @rmbolger, do you have any timescales for the next release that will incorporate this PR?

Thanks :)

rmbolger:
There are two other open issues that I've committed fixes for, 157 and 168. I was hoping to get confirmation from the OPs on those issues that the changes do indeed fix their issue before pushing the release. But if I don't hear back in a few days, I'll just push it anyway. So likely before this weekend.

matt-FFFFFF:
Fantastic. Thanks for the prompt response.

PR #173, opened 2019-09-16
Azure App Services / Function Apps Storage Path

alphaz18:
In Azure App Services and Function Apps, the LOCALAPPDATA folder is a non persistent storage location. that means it will randomly get wiped by Microsoft. This doesn't work for storing certs and metadata. so they recommend using the HOME env var folder as this is the only persistent storage in App services. 
This code will determine if its on azure app services by using (env var WEBSITE_INSTANCE_ID) which only exists in those services.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 .vscode
+/.vs/Posh-ACME/v16
+/.vs
--- a/Posh-ACME/Private/Import-PAConfig.ps1
+++ b/Posh-ACME/Private/Import-PAConfig.ps1
@@ -26,8 +26,13 @@ function Import-PAConfig {
     # make sure we have the root config folder
     if ([string]::IsNullOrWhiteSpace((Get-ConfigRoot))) {
         if ($IsWindows -or $PSVersionTable.PSEdition -eq 'Desktop') {
-            Set-ConfigRoot (Join-Path $env:LOCALAPPDATA 'Posh-ACME')
-        } elseif ($IsLinux) {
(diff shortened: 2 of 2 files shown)


rmbolger:
Hey @alphaz18. Thanks for the idea! However, technically this would be a breaking change for anyone using Posh-ACME on these platforms that is currently relying on the fact that LOCALAPPDATA is non-persistent. I can definitely see some use-cases where it may even be desirable for short-lived services or instances. So I wouldn't want to force this change on everyone.

As an alternative if you want your own instances to have a persistent config location, you should be able to use the `POSHACME_HOME` environment variable to override the default config location to wherever you want. It looks like you'd configure this in [Application settings](https://docs.microsoft.com/en-us/azure/app-service/configure-common). Would that work?

alphaz18:
thanks, let me give that a try! I"m ok with that, but I'm just curious, under what scenario would you want data like that to randomly disappear? the data is only persistent over in that one webapp or one function app, if the function app is deleted, then the data will be gone anyway? (not being combative, just genuinely curious)

rmbolger:
I was thinking micro-services where each instance has a unique (potentially random) DNS name such that the certificate is only good once and the service's lifetime is relatively short. But I don't regularly play in that space, so you're probably right that it doesn't matter whether the config is persistent or not in that case.

Mostly, I'm just trying to avoid breaking changes (as minor as this one might be) unless it's a really good reason. It also feels weird that it's core module code tailored specifically for Azure.

PR #183, opened 2019-10-12
Unoeuro plugin

OrKarstoft:
Hi,

I've created a Unoeuro plugin for Posh-ACME. Wonderful module. I use this Unoeuro plugin for my domains and wish to share.

Tested with single, multi and wildcart domains.

### DIFF
--- a/Posh-ACME/DnsPlugins/Unoeuro-readme.md
+++ b/Posh-ACME/DnsPlugins/Unoeuro-readme.md
@@ -1,0 +1,15 @@
+# How To Use the Unoeuro DNS Plugin
+
+This plugin works against the [Unoeuro](https://www.unoeuro.com/dns) DNS provider. It is assumed that you have already setup an account and created the DNS zone(s) you will be working against.
+
+## Finding your information in Unoeuro
+
+Using Unoeuro API requires only your account name or account number and API Key which can be found on the [account](https://www.unoeuro.com/en/controlpanel/account/) page.
+
+
+## Using the Plugin
+
+```powershell
+$pArgs = @{ UEAccount='UE123456'; UEAPIKey='ABCDEFghijkLmNoPq' }
+New-PACertificate example.com -DnsPlugin unoeuro -PluginArgs $pArgs
+```
--- a/Posh-ACME/DnsPlugins/Unoeuro.ps1
+++ b/Posh-ACME/DnsPlugins/Unoeuro.ps1
@@ -35,7 +35,7 @@ function Add-DnsTxtUnoEuro {
     # check for an existing record
     Write-Verbose "[UE-Plugin] Running: GET $UEApiRoot/$UEAccount/$UEAPIKey/my/products/$UEdomain/dns/records/"
     try {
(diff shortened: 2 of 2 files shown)


rmbolger:
Hey @OrKarstoft. This is great! Thank you.

My only concern at the moment is the regex for determining the apex domain from the record name. At the moment, it will break if the user puts capital letters in the domain name or uses one of the newer gTLDs that are longer than 3 characters (.store, .site, .email, etc). It's going to be hard to get a reliable regex that is future proofed against the ever-changing TLD landscape.

I'd suggest following the model a lot of other plugins use where you basically break the record name down into smaller and smaller pieces split on the `.` and then query the API for that zone to see if it exists. The `GET /my/products/[object]/dns` endpoint would likely work for this. Check out the `Find-GDZone` function in the GoDaddy plugin for an example.

Once you have the apex zone, you can then split the short record name from it like this.

```powershell
$recShort = $RecordName -ireplace [regex]::Escape(".$zone"), [string]::Empty
```

There have been plugin bugs in the past where the user entered weird capitalization that screwed up the process of splitting the record name from the domain. This seems to be a fairly reliable way around those issues.

I'm also curious whether unoeuro's api is case sensitive or not. Like, if my domain is `example.com` and I do a `GET /my/products/ExAmPle.CoM/dns`, will it succeed or fail?

OrKarstoft:
Hey @rmbolger,

Thank you very much for the excellent feedback! I share your concern and will start working on this right away.

I do admit that I did not think of the newer gTLDs that are longer than three characters nor did I think of anyone writing their domain in other than lowercase. Therefore i fully understand that I need to change the way the plugin searches for the domain. Thank you very much for the examples and great explanation in your comment.

I'll investigate the Unoeuro API to answer your question about case sensitivity.

OrKarstoft:
Hey @rmbolger,

First I'd like to answer your question if the Unoeuro API is case sensitive. I've tested with my own domain using `GET`. Running `https://api.unoeuro.com/1/UE123456/APIKEY/my/products/eXamPlE.com/dns/records/` does return code 400. As does `https://api.unoeuro.com/1/UE123456/APIKEY/mY/ProDucTs/example.com/dns/records/`. I've made multiple tests with the unoeuro plugin with different domains and subdomains containing capital letters in random order. It works with no problems as you can see on the following pictures.

Following picture shows the certificate generated. Looks like a normal certificate with all lowercase domain and subdomain.
CapitalCert

Here's the domain and subdomain entered for generating the previous certificate.
capitalCertOrder

Another example is this;
Running following command with the domain `test1.TeSt2.exaMPle.com`.
CapitalCertPS
Generated following record in Unoeuro.
CapitalCertUnoeuro
And this certificate
CapitalCert2

Tests with capital TLD (.dK) also OK!

This should prove that the certificate comes out as expected even though someone would type in the domain in weird ways.


As for the regex. I've implemented the solution you proposed. The GoDaddy function `Find-GDZone` worked as a charm with a little modification to fit the needs of Unoeuro.


Looking forward to hear from you.
Best regards

rmbolger:
Awesome. Thanks for the effort. Merging now.

Issue #184, opened 2019-10-13
Cloudflare subdomain issue


mikopia:
I am trying to generate a certificate for a sub domain of a domain i have in Cloudflare.

Domain: domain.com
Subdomain: app1.domain.com

The problem is that the cloudflare plugin can't find the zone called app1.domain.com, so it fails.

```
PS C:\Users\Administrator> New-PACertificate app1.domain.com
 -DnsPlugin Cloudflare -PluginArgs $pArgs
Unable to find Cloudflare hosted zone for _acme-challenge.app1.domain.com
At C:\Users\Administrator\Documents\WindowsPowerShell\Modules\Posh-ACME\3.8.0\DnsPlugins\Cloudflare.ps1:103 char:9
+         throw "Unable to find Cloudflare hosted zone for $RecordName"
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (Unable to find app1.domain.com
:String) [], RuntimeException
    + FullyQualifiedErrorId : Unable to find Cloudflare hosted zone for _acme-challenge.app1.domain.com

```
I overrode this by specifying a zoneID in cloudflare.ps1 and the rest worked.

rmbolger:
Which type of credential are you using?

MetUys:
Odd I had no issue with this an even deeper use... app1.sub1.domain.ccTLD
my code:
`New-PACertificate app1.sub1.domain.ccTLD -AcceptTOS -Contact $Email -DnsPlugin Cloudflare -PluginArgs @{CFAuthEmail=$Email; CFAuthKey='ohnoyoudont'} -PfxPass $CertsPassword -DnsSleep 5 -Install -Verbose`

Jackbennett:
With the api token set to specifically my example.co.uk zone Find-CFzone errors;

```powershell
Invoke-RestMethod : The remote server returned an error: (403) Forbidden.
At C:\Program Files\WindowsPowerShell\Modules\posh-acme\3.8.0\DnsPlugins\CloudFlare.ps1:253 char:27
+ ...    $result = Invoke-RestMethod "$($apiRoot)?page=$page&per_page=50"
```

If I hit `https://api.cloudflare.com/client/v4/zones` manually when my API token is set to a specific zone I get a 403

[...]

PR #186, opened 2019-10-15
If bLock skips even when script variable does not exist.

Jackbennett:
When debugging #184 I found the block for getting zone records from cloudflare was being completely skipped. Then `$script:CFZoneIDs.ContainsKey($zoneTest)` never returns because it's and empty list of zones.  You can see "checking co.uk" in the debug logs as it fails through the loop.

I'm not clear why `-not $script:CFZoneIDs` doesn't. I presume something about scoping the variable makes it harder to test.


rmbolger:
I've reverted the change that was using `$script:CFZoneIDs`, so this is no longer relevant.

PR #218, opened 2020-03-05
Adds Yandex plugin

WiZaRd31337:
Adds support for DNS zones hosted at Yandex (https://connect.yandex.com/)

### DIFF
--- a/Posh-ACME/DnsPlugins/Yandex-Readme.md
+++ b/Posh-ACME/DnsPlugins/Yandex-Readme.md
@@ -11,6 +11,7 @@ You will need the "PDD" administrator token.
 ## Using the Plugin
 
 The plugin arguments you need is the administrator token created earlier and domain name delegated on yandex.
+Also, you should use DNSSleep parameter more than 900, bacause 15 min is a minimal allowed ttl value for Yandex.
 
 ```powershell
 New-PACertificate test.domain.zone -DnsPlugin Yandex -PluginArgs @{YandexApiKey='xxxxxxxxxxxxxxxx'; DomainName='domain.zone'} -DnsSleep 1000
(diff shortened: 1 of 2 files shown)


rmbolger:
Hey @WiZaRd31337. Thanks for the new plugin! I had been meaning to add Yandex support for a while and never got around to it. So yay! I'll try to review it soon and get it merged.

rmbolger:
Hi @WiZaRd31337. There are a few problems I'm seeing with the current implementation at the moment.

The `DomainName` argument shouldn't be necessary. Most plugins use the `RecordName` parameter to determine the domain by querying the provider's API for subsets of the RecordName FQDN. Check some of the other plugins for their `Find-BlahZone` function to see how they typically do it. I might mess around with implementing this as well, I just need to find the yandex API docs.

Another thing I'm seeing is if you run `Publish-DnsChallenge` against the same record for two different token values, the second one overwrites the first one's record instead of adding a second record. The plugin needs to be able to add multiple TXT records for the same name to support wildcard certs. It's the same idea as adding multiple A records for the same FQDN to do DNS round robin.

Finally, the Remove function should silently succeed if it is asked to remove a record that no longer exists. At the moment, it throws a "bad_record_id" error if you run the same `Unpublish-DnsChallenge` command twice.

rmbolger:
I hope you don't mind, I went a little crazy and made some significant changes to address the issues I brought up in my previous comment. I also tweaked some things so it acts more like other plugins. In particular, `YandexApiKey` is now either `YDAdminToken` or `YDAdminTokenInsecure`. You also no longer need to specify the `DomainName` parameter because it will find it based on the record name.

Everything seems to be working for me. But I'd love to get confirmation that it works for you before merging it.

WiZaRd31337:
Hi @rmbolger. Thank you for your review and fixes! I will check shortly. Also, I will fix the code for my second PR, because it contains the same errors)

rmbolger:
> Finally, the Remove function should silently succeed if it is asked to remove a record that no longer exists. At the moment, it throws a "bad_record_id" error if you run the same Unpublish-DnsChallenge command twice.

Similarly, errors should also be ignored in the Add function if the record already exists with the correct value.

WiZaRd31337:
I confirm, now all works as expected. Thank you so much! And I am still working on my second PR)

PR #226, opened 2020-03-13
Adds dnspod.com plugin

WiZaRd31337:
...And one more plugin, yes)


rmbolger:
You're crazy in a good way. I can't even keep up. What's funny about this provider is that I already have an account with them that I forgot about. It's like I was going to write a plugin myself and forgot.

rmbolger:
It looks like a lot, but the vast majority of my changes were cosmetic. I added `-EA Stop` and `@script:UseBasic` to all the Invoke-RestMethod calls. The ErrorAction stop is useful because Invoke-RestMethod doesn't throw a terminating error by default on HTTP error codes, so the try/catch won't catch it. The `@script:UseBasic` is important for compatibility but easy to forget.

I also fixed a bug I had because my password has special characters that apparently don't transfer well to the login endpoint unless they're URL encoded.

In any case, nice work. Ready to merge.

PR #232, opened 2020-03-23
Fix Insecure/Non-windows key parameter example

webprofusion-chrisc:
Update docs to correct the parameter name for the non-secure version of the key


rmbolger:
Nice catch! Thanks.

PR #233, opened 2020-03-24
Non-windows should use the LITokenInsecure parameter

webprofusion-chrisc:
Fix non-secure parameter reference


rmbolger:
Thanks again.

PR #239, opened 2020-04-14
DNS Plugin for Hetzner.de

derguterat:
# How To Use the Hetzner DNS Plugin

This plugin works against the [Hetzner](https://www.hetzner.de/) DNS provider. It is assumed that you have already setup an account and created the DNS zone(s) you will be working against.

## Finding your information in Hetzner

Using Hetzner DNS API requires only your account name or account number and Auth-API-Token which can be found on the [account](https://dns.hetzner.com/) page.

## Using the Plugin

```powershell
$pArgs = @{ HetznerAPIKey='yyyyyyyy' }
New-PACertificate example.com -DnsPlugin Hetzner -PluginArgs $pArgs
```

### DIFF
--- a/Posh-ACME/DnsPlugins/Hetzner-Readme.md
+++ b/Posh-ACME/DnsPlugins/Hetzner-Readme.md
@@ -1,0 +1,26 @@
+# How To Use the Hetzner DNS Plugin
+
+This plugin works against the [Hetzner](https://www.hetzner.de/) DNS provider. It is assumed that you have already setup an account and created the DNS zone(s) you will be working against.
+
+## Setup
+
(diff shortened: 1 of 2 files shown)


rmbolger:
Hey there @derguterat. Good job submitting the pull request!

I took the liberty of tweaking your readme content and adding it to an actual file in the PR. I also fixed a bug that I ran into where if you try to add multiple records for the same name with different values, it was overwriting the previous one instead of adding a second one. I did some general code cleanup as well. But overall, it seems to be working for me. Can you verify the changes I made didn't break anything for you?

Also, do you know what the deal is with the `X-Consumer-Username` header? In their API docs, the curl examples don't include it, but the rest of them do. And the ones that do all include it as an empty string which seems silly. The API doesn't seem to care either way.

derguterat:
Hey there @rmbolger. Many thanks for the revision of the code. I verifed the changes, it works.

I made a request to Hetzner support and received the answer that the header "X-Consumer-Username" is no longer needed. Because of this information I have removed the additional header.

rmbolger:
Sorry for the delay on the merge. Got busy and totally slipped my mind. Thanks again for the effort!

PR #245, opened 2020-05-03
DNS Plugin: Domain Offensive

Armitxes:
Thanks for your work. Thought I share my DNS Plugin with you that I use for my hoster [Domain Offensive](https://www.do.de/). Hope I didn't oversee anything but if so, feel free to modify it.

# How To Use the Domain Offensive DNS Plugin

This plugin works against the [Domain Offensive](https://www.do.de/) DNS provider. It is assumed that you have already setup an account and created the DNS zone(s) you will be working against.

## Setup

We need to retrieve an secret API token for the account that will be used to update DNS records. Further information can ba found at [the (german) developer docs.](https://www.do.de/wiki/LetsEncrypt_-_Entwickler).

## Using the Plugin

Your personal API token is specified using the `DomOffToken` or `DomOffTokenInsecure` parameter.

### Windows and/or PS 6.2+ only (secure string)
```powershell
$pArgs = @{
    DomOffToken = $secureString
}
New-PACertificate example.com -DnsPlugin DomainOffensive -PluginArgs $pArgs
```

### Any OS (default string)
```powershell
$pArgs = @{
    DomOffTokenInsecure = '1md6xRcUCTrB58kbpwAH'
}
New-PACertificate example.com -DnsPlugin DomainOffensive -PluginArgs $pArgs
```

### DIFF
--- a/Posh-ACME/DnsPlugins/DomainOffensive-Readme.md
+++ b/Posh-ACME/DnsPlugins/DomainOffensive-Readme.md
@@ -1,0 +1,22 @@
+# How To Use the Domain Offensive DNS Plugin
+
+This plugin works against the [Domain Offensive](https://www.do.de/) DNS provider. It is assumed that you have already setup an account and created the DNS zone(s) you will be working against.
+
+## Setup
+
+We need to retrieve an secret API token for the account that will be used to update DNS records. Further information can ba found at [the (german) developer docs.](https://www.do.de/wiki/LetsEncrypt_-_Entwickler).
+
+## Using the Plugin
+
+Your personal API token is specified using the generated `Token` parameter.
+The fully qualified name for the TXT record can be specified using the `RecordName` parameter.
+The value for the TXT record can be specified using the `TxtValue` parameter.
+
+```powershell
+$pArgs = @{
+    Token = '1md6xRcUCTrB58kbpwAH',
+    Domain = '_acme-challenge.site1.example.com',
+    TxtValue = 'OVxwaDm7MgN1IRG0eSivJMlepO9CL4X8vKo6Tcns'
+}
+New-PACertificate example.com -DnsPlugin DomainOffensive -PluginArgs $pArgs
+```
--- a/Posh-ACME/DnsPlugins/DomainOffensive-Readme.md
+++ b/Posh-ACME/DnsPlugins/DomainOffensive-Readme.md
@@ -11,8 +11,15 @@ We need to retrieve an secret API token for the account that will be used to upd
 Your personal API token is specified using the `Token` parameter.
 
 ```powershell
+# Example: Provide the private token via secure string (Windows or any OS with PowerShell 6.2+)
 $pArgs = @{
-    Token = '1md6xRcUCTrB58kbpwAH'
+    DomOffToken = $secureString
+}
+New-PACertificate example.com -DnsPlugin DomainOffensive -PluginArgs $pArgs
+
+# Alternative Example by directly providing the private token as string
+$pArgs = @{
+    DomOffTokenInsecure = '1md6xRcUCTrB58kbpwAH'
 }
 New-PACertificate example.com -DnsPlugin DomainOffensive -PluginArgs $pArgs
 ```
--- a/Posh-ACME/DnsPlugins/DomainOffensive-Readme.md
+++ b/Posh-ACME/DnsPlugins/DomainOffensive-Readme.md
@@ -8,15 +8,11 @@ We need to retrieve an secret API token for the account that will be used to upd
 
 ## Using the Plugin
 
-Your personal API token is specified using the generated `Token` parameter.
-The fully qualified name for the TXT record can be specified using the `RecordName` parameter.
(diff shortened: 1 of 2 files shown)


rmbolger:
Hey @Armitxes, thanks for the plugin! This is the first provider I've seen that has an API specifically for Let's Encrypt.

The only problem I see at the moment is that the parameter blocks in the functions aren't quite right. The `$RecordName` and `$TxtValue` params should remain as the first and second in both the add and remove functions. Both also need to still have the `$ExtraParams` param in there at the end as well. This allows people to use multiple different plugins for names in a single cert.

I would also rename the `$Token` param to something more specific to Domain Offensive like `$DomOffToken`. Each plugin's parameters need to be unique among all plugins.

If you want to make things even better, take a look at how some of the other plugins offer a "secure" and "insecure" version of things like secret tokens. This allows them to have their secret values encrypted when they're saved on the local drive.

Posh-ACME/DnsPlugins/DomainOffensive-Readme.md
```diff
@@ -0,0 +1,22 @@
...
+
+Your personal API token is specified using the generated `Token` parameter.
+The fully qualified name for the TXT record can be specified using the `RecordName` parameter.
+The value for the TXT record can be specified using the `TxtValue` parameter.
+
+```powershell
+$pArgs = @{
+    Token = '1md6xRcUCTrB58kbpwAH',
+    Domain = '_acme-challenge.site1.example.com',
+    TxtValue = 'OVxwaDm7MgN1IRG0eSivJMlepO9CL4X8vKo6Tcns'
```
rmbolger, line 19: These shouldn't be necessary to pass as plugin parameters. They're automatically passed by the module to the plugin.

Posh-ACME/DnsPlugins/DomainOffensive.ps1
```diff
@@ -0,0 +1,103 @@
+function Add-DnsTxtDomainOffensive {
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory,Position=0)]
+        [string]$Token,
+        [Parameter(Mandatory,Position=1)]
+        [string]$RecordName,
+        [Parameter(Mandatory,Position=2)]
+        [string]$TxtValue
```
rmbolger, line 9: These need to be positional params 0 and 1. And the `$ExtraParams` one needs to be added back at the end of the param block.

Posh-ACME/DnsPlugins/DomainOffensive.ps1
```diff
@@ -0,0 +1,103 @@
...
+    #>
+}
+
+function Remove-DnsTxtDomainOffensive {
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory,Position=0)]
+        [string]$Token,
+        [Parameter(Mandatory,Position=1)]
+        [string]$RecordName
```
rmbolger, line 52: This param block should be identical to the Add one even if you don't need all of the values.

Posh-ACME/DnsPlugins/DomainOffensive.ps1
```diff
@@ -0,0 +1,103 @@
...
+    .PARAMETER Token
+        Token provided by Domain Offensive.
+
+    .PARAMETER Domain
+        The fully qualified name of the TXT record to be removed.
+
+    .EXAMPLE
+        Add-DnsTxtDomainOffensive '1md6xRcUCTrB58kbpwAH' '_acme-challenge.site1.example.com'
+
+        Adds a TXT record for the specified site with the specified value using the account associated with the given token.
```
rmbolger, line 79: I think you forgot to update this example after copying it from the Add function.

Posh-ACME/DnsPlugins/DomainOffensive-Readme.md
Armitxes: Good to know. Made the mistake to pass the parameters myself.
Fixed.

Posh-ACME/DnsPlugins/DomainOffensive.ps1
Armitxes: Fixed order and added the `$ExtraParams`

Posh-ACME/DnsPlugins/DomainOffensive.ps1
Armitxes: Done

Posh-ACME/DnsPlugins/DomainOffensive.ps1
Armitxes: Yep, did a little Oopsie there.. fixed.
Thanks for reviewing :)

Armitxes:
@rmbolger Thanks for your review. I think I covered all of your suggestions. Tested with secure string and it worked aswell.

Let me know if there is anything else to do.

rmbolger:
Awesome. Thanks for the updates. Made some tiny additional tweaks with the help stuff. But otherwise, this is good to go.

Issue #258, opened 2020-06-16
Add support for external account binding


rmbolger:
Some non-Let's Encrypt CAs that implement ACME require the External Account Binding feature described in [RFC 8555 section 7.3.4](https://tools.ietf.org/html/rfc8555#section-7.3.4).

Posh-ACME does not currently support any way to provide the key identifier or MAC key needed for creating a new ACME account on these servers and thus can't be used with them.

PR #259, opened 2020-06-16
Add External Account Binding support

rmbolger:
This change addresses #258 and adds additional parameters to `New-PAAccount` that support the external account binding feature described in [RFC 8555 section 7.3.4](https://tools.ietf.org/html/rfc8555#section-7.3.4).

- `-ExtAccountKID` accepts a string value for the key identifier provided by the CA
- `-ExtAccountHMACKey` accepts a string value provided by the CA that represents a Base64Url encoded HMAC key.
- `-ExtAccountAlgorithm` accepts a string value from the set `HS256`, `HS384`,`HS512` which indicates the algorithm to use for creating the MAC from the key. The default value is HS256.

The internal `New-Jws` function was modified to support HMAC keys for JWS generation.

The new parameters in `New-PAAccount` were not also exposed in `New-PACertificate`. Users must explicitly create their ACME account with external binding prior to running `New-PACertificate`.

`New-PAAccount` also now recognizes when an ACME server requires external account binding and provides a better error message to the end user if they haven't supplied the required parameters.

### DIFF
--- a/Posh-ACME/Private/New-Jws.ps1
+++ b/Posh-ACME/Private/New-Jws.ps1
@@ -103,6 +103,7 @@ function New-Jws {
         # create the signature
         $HashAlgo = [Security.Cryptography.HashAlgorithmName]::SHA256
         $PaddingType = [Security.Cryptography.RSASignaturePadding]::Pkcs1
+        Write-Debug "Signing message using $($Header.alg)"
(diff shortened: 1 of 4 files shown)


rmbolger:
Need to add help content for the additional parameters before merging this.

PR #266, opened 2020-07-27
PluginArgs for Submit-Renewal -AllOrders & -AllAccounts

f-bader:
Currently the parameter "PluginArgs" in Submit-Renewal only works with ParameterSet 'Specific'.
This change modifies 'AllOrders' and 'AllAccounts to use the PluginArgs if present.


rmbolger:
Thanks @f-bader. I had totally forgotten about this use-case. After reviewing the code a bit, it looks like the `-NoSkipManualDns` switch param should be added as well. Do you want to add that before I merge?

f-bader:
Sure, will do

f-bader:
@rmbolger All done, please check if this is fine with you

rmbolger:
Perfect. Thanks again.

PR #276, opened 2020-09-02
Allow authorization objects without the expires property set

mortenmw:
Posh-ACME will currently fail if a CA does not return expires time in the authorization object. RFC 8555 7.1.4 does not require the expire time to be set unless the status property is set to "valid". For all other statuses the timestamp is optional. 

An example of an RFC 8555 compliant CA returning auth objects without the expiration is Buypass (https://api.buypass.com/acme/directory and https://api.test4.buypass.no/acme/directory). 

This change fixes it by ignoring the expires property if it does not exist instead of stopping with an exception message.

### DIFF
--- a/Posh-ACME/Public/Get-PAAuthorizations.ps1
+++ b/Posh-ACME/Public/Get-PAAuthorizations.ps1
@@ -54,9 +54,6 @@ function Get-PAAuthorizations {
             $auth.PSObject.TypeNames.Insert(0,'PoshACME.PAAuthorization')
             Write-Debug "Response: $($auth | ConvertTo-Json)"
 
-            # fix any dates that may have been parsed by PSCore's JSON serializer
-            $auth.expires = Repair-ISODate $auth.expires
-
             # Workaround non-compliant ACME servers such as Nexus CM that don't include
             # the status field on challenge objects. Just copy the auth's status to
             # each challenge.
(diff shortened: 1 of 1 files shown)


rmbolger:
Thanks, @mortenmw! This is a great find. I have a small implementation suggestion I'm going to try and post in-line.

Posh-ACME/Public/Get-PAAuthorizations.ps1
```diff
@@ -71,6 +68,14 @@ function Get-PAAuthorizations {
                 Write-Warning "ACME server returned non-compliant challenge objects with no status. Please report this to your ACME server vendor."
             }
 
+            # According to RFC 8555 7.1.4 the expires property is only REQUIRED when the property status is "valid".
+            # It's OPTIONAL for any other status and some CA's will not return it.
+            # Only repair the timestamp if it actually exists
+            if([bool]($auth.PSobject.Properties.name -match "expires")) {
```
rmbolger, line 74: Rather than using Regex via the `-match` operator, I'd prefer to check the existence of the property similar to how it checks for status in the challenges just above this. So:

```powershell
if ('expires' -in $auth.PSObject.Properties.Name) {
```

mortenmw: I agree, thats a cleaner and better way of doing it

rmbolger: In case you didn't already know, you should be able to commit the change to your local copy and push the change up to your fork on Github and it will show up here without needing to submit a new pull request.

mortenmw: Thanks, its commited now

rmbolger:
Thanks again for the fix.

PR #277, opened 2020-09-03
Update Tutorial.md

Leon99:
Fix syntax error


rmbolger:
Thanks @Leon99. Good catch.

Issue #289, opened 2020-11-19
remove query parameters if returned in providers response location 


KaiWalter:
I want to work with a provider where directory URL requires additional an query parameter like:

`https://some.service.provider.net/acme/acct/-AbcdefghijK?configurationId=test`

this results in an error in 'New-PAAccount'

```
Out-File: C:\Users\kai\src\Posh-ACME\Posh-ACME\Public\New-PAAccount.ps1:185
Line |
 185 |  … rtTo-Json | Out-File (Join-Path $script:AcctFolder 'acct.json') -Forc …
     |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Cannot perform operation because the wildcard path C:\Users\xxxxx\AppData\Local\Posh-ACME\some.service.provider.net\-AbcdefghijK?configurationId=test\acct.json did not resolve to
     | a file.
```

Along with this issue I would like to submit a small PR to fix this.

rmbolger:
Thanks @KaiWalter. Looks like my comment prediction came true. Doh! https://github.com/rmbolger/Posh-ACME/blob/1a8b8373c8f7aa6322b5787eba489b1995737e10/Posh-ACME/Public/New-PAAccount.ps1#L145-L150

Out of curiosity which ACME provider is this? And do you think the provider's dependence on query parameters will cascade into other issues like the module's assumption that each ACME server has a unique hostname (that gets translated into a folder name in the config)?

KaiWalter:
It is an ACME service from our on-prem service provider Atos.

rmbolger:
Fix merged. Thanks again!

PR #290, opened 2020-11-19
remove query parameters from location header response

KaiWalter:
this PR removes trailing query parameters for issue #289


KaiWalter:
You're right - parsing it actually with Uri is probably more consistent and nicer.

rmbolger (approved):

PR #292, opened 2020-11-23
Add support for different Azure cloud environments

InKahootz:
Fixes:  #293

### DIFF
--- a/Posh-ACME/DnsPlugins/Azure-Readme.md
+++ b/Posh-ACME/DnsPlugins/Azure-Readme.md
@@ -295,6 +295,20 @@ $azParams = @{
 New-PACertificate test.example.com -DnsPlugin Azure -PluginArgs $azParams
 ```
 
+### Azure Cloud Environments
+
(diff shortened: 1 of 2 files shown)


InKahootz:
Any feedback is welcome 😄

rmbolger:
Hah, you beat me to the validationset fix by minutes. I made a small tweak to the usage guide so the AZEnvironment stuff is more prominent in the Using section.

rmbolger:
In any case, everything is working for me. Stylistically, I have some minor hesitations with the AzureCloud class definition due to some legacy hassles I've had with PowerShell classes. I know things *should* work since everyone using the module should at least be on 5.1. But I'd probably have just defined those as plain old hashtable objects.

InKahootz:
I refactored to hashtable so there wouldn't be any worry about classes. It's a drop in replacement so nothing big needed to be changed.

rmbolger (approved):

rmbolger:
Love it. Merging.

InKahootz:
Thanks for the merge and 3.20.0 release. Already using in our production pipeline and it works great!

PR #309, opened 2021-01-20
DNS Plugin for Infomaniak.com

Sundypha:
Hi @rmbloger. 
First of all: Thank you so much for this ACME client.

I added a DNS Plugin for the hoster Infomaniak from Switzerland. 

I did use the plugin on two Windows Servers and a Windows Desktop which all worked well.
I tested:
- example.com
- *.example.com
- sub.example.com
- sub2.sub1.example.com

I've got one question. Is there a way to tell the user if something failed? e.g., the zone hasn't been found? Refer to line [69](https://github.com/Sundypha/Posh-ACME/blob/dns1-infomaniak-plugin/Posh-ACME/Plugins/Infomaniak.ps1#L69) and line [150](https://github.com/Sundypha/Posh-ACME/blob/dns1-infomaniak-plugin/Posh-ACME/Plugins/Infomaniak.ps1#L150) where I could fail gracefully.

Awaiting your review and feedback!

Best
Simon

### DIFF
--- a/Posh-ACME/Plugins/Infomaniak.ps1
+++ b/Posh-ACME/Plugins/Infomaniak.ps1
@@ -70,6 +70,9 @@ Function Add-DnsTxt {
             {
                 Write-Verbose "Record $RecordName added with value $TxtValue."
             }
+            {
+                throw "Record $RecordName with value $TxtValue could not be added."
+            }
         } catch { throw }
     }
 
@@ -150,6 +153,10 @@ Function Remove-DnsTxt {
             if($response.result -eq 'success')
             {
                 Write-Verbose "Record $RecordName deleted."
+            }
+            else
+            {
+                throw "Record $RecordName could not be deleted."
             }
         } catch { throw }
     } else {
--- a/Posh-ACME/Plugins/Infomaniak.ps1
+++ b/Posh-ACME/Plugins/Infomaniak.ps1
@@ -47,8 +47,8 @@ Function Add-DnsTxt {
     # check for a matching record
     $rec = $recs.data | Where-Object {
         $_.type -eq 'TXT' -and
-        $_.source -eq $recShort -and
-        $_.target -eq $TxtValue
+        $_.source_idn -eq $RecordName -and
+        $_.target_idn -eq $TxtValue
     }
 
     if ($rec) {
--- a/Posh-ACME/Plugins/Infomaniak.ps1
+++ b/Posh-ACME/Plugins/Infomaniak.ps1
@@ -216,6 +216,8 @@ Function Find-InfomaniakZone {
         [hashtable]$RestParameters
     )
 
+    $apiRoot = 'https://api.infomaniak.com'
+
     # setup a module variable to cache the record to zone mapping
     # so it's quicker to find later
     if (!$script:InfomaniakRecordZones) { $script:InfomaniakRecordZones = @{} }
--- a/Posh-ACME/Plugins/Infomaniak.ps1
+++ b/Posh-ACME/Plugins/Infomaniak.ps1
@@ -226,10 +226,7 @@ Function Find-InfomaniakZone {
     # - sub1.sub2.example.com
     # - sub2.example.com
     # - example.com
-
-    # remove _acme-challenge. from requested RecordName to get the zone
-    $zone = $RecordName.Split('.',2)[1];
-    $zoneTest = $zone;
+    $zoneTest = $RecordName;
     while($zoneTest.Contains('.'))
     {
         Write-Debug "Checking $zoneTest"
@@ -251,6 +248,8 @@ Function Find-InfomaniakZone {
         $zoneTest = $zoneTest.Split('.',2)[1]
     }
 
+    # remove _acme-challenge. from requested zone
+    $zone = $RecordName.Split('.',2)[1]
     Write-Debug "Zone $zone does not exist ..."
     return $null
 }
--- a/Posh-ACME/Plugins/Infomaniak.ps1
+++ b/Posh-ACME/Plugins/Infomaniak.ps1
@@ -226,7 +226,10 @@ Function Find-InfomaniakZone {
     # - sub1.sub2.example.com
     # - sub2.example.com
     # - example.com
-    $zoneTest = $RecordName;
+
+    # remove _acme-challenge. from requested RecordName to get the zone
+    $zone = $RecordName.Split('.',2)[1];
+    $zoneTest = $zone;
     while($zoneTest.Contains('.'))
     {
         Write-Debug "Checking $zoneTest"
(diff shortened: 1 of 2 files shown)


Sundypha:
Just figured that in the function `Find-InfomaniakZone` I query the `zone _acme-challenge.example.com` which is unnecessary. Going to create a commit shortly.

rmbolger:
Hi @Sundypha. Thanks for the submission! Everything looks great at first glance. Give me a bit to review and test.

Posh-ACME/Plugins/Infomaniak.ps1
```diff
@@ -0,0 +1,256 @@
...
+    # sub-zone that would hold the record rather than just adding it to the apex. So for something
+    # like _acme-challenge.site1.sub1.sub2.example.com, we'd look for zone matches in the following
+    # order:
+    # - site1.sub1.sub2.example.com
+    # - sub1.sub2.example.com
+    # - sub2.example.com
+    # - example.com
+
+    # remove _acme-challenge. from requested RecordName to get the zone
+    $zone = $RecordName.Split('.',2)[1];
```
rmbolger, line 231: While this might seem like an easy optimization at first, it can actually break things for people using CNAME aliases for challenges. The record being created may not always be `_acme-challenge.blah`, so you have to check the whole thing first even if most of the time it will fail.

rmbolger (changes requested):
Great job. Just fix the one _acme-challenge check removal and this should be good to merge.

Posh-ACME/Plugins/Infomaniak.ps1
Sundypha: Thanks. Will try with an alias.

rmbolger:
> I've got one question. Is there a way to tell the user if something failed? e.g., the zone hasn't been found? Refer to line 69 and line 150 where I could fail gracefully.

Forgot to answer this. For things like the domain not being found which you can't really recover from, I'll just use `throw "my failure message"` to send an exception back up to the user so they can decide what to do.

Posh-ACME/Plugins/Infomaniak.ps1
rmbolger: Specifically, try running the following which is what can theoretically happen when people point an alias to the root of a domain:

```powershell
Publish-Challenge _acme-challenge.example.com (Get-PAAccount) faketoken Infomaniak $pArgs -DnsAlias aliasdomain.com -Verbose
```

In this case `aliasdomain.com` is what gets sent to the plugin in the `$RecordName` variable.

Sundypha:
You were right. It failed when I had alias `abc.example.com` since it was only checking the zone `example.com` instead of the `abc` zone. Reverted the commit and added `throw`s.

Posh-ACME/Plugins/Infomaniak.ps1
```diff
@@ -0,0 +1,262 @@
...
+        try {
+            $response = Invoke-RestMethod "$apiRoot/1/domain/$($zone.id)/dns/record" -Method Post -Body $json `
+                @restParams @Script:UseBasic -EA Stop
+            if($response.result -eq 'success')
+            {
+                Write-Verbose "Record $RecordName added with value $TxtValue."
+            }
+            {
+                throw "Record $RecordName with value $TxtValue could not be added."
+            }
```
rmbolger, line 75: I think you're missing an `else` in here.

Posh-ACME/Plugins/Infomaniak.ps1
```diff
@@ -0,0 +1,262 @@
...
+                return $script:InfomaniakRecordZones.$RecordName;
+            }
+        } catch { throw }
+        # remove one sub site
+        $zoneTest = $zoneTest.Split('.',2)[1]
+    }
+
+    # remove _acme-challenge. from requested zone
+    $zone = $RecordName.Split('.',2)[1]
+    Write-Debug "Zone $zone does not exist ..."
```
rmbolger, line 260: Another remnant of removing `_acme-challenge` to cleanup. You might just change the debug message to something like 

```powershell
Write-Debug "Zone for $RecordName not found..."
```

Posh-ACME/Plugins/Infomaniak.ps1
Sundypha: yes. Strange that during Publish-Challenge PS didn't throw an error.

Posh-ACME/Plugins/Infomaniak.ps1
Sundypha: Agree. Changed that for coherency.

Posh-ACME/Plugins/Infomaniak.ps1
```diff
@@ -255,8 +256,6 @@ Function Find-InfomaniakZone {
         $zoneTest = $zoneTest.Split('.',2)[1]
     }
 
-    # remove _acme-challenge. from requested zone
-    $zone = $RecordName.Split('.',2)[1]
     Write-Debug "Zone $zone does not exist ..."
```
rmbolger, line 259: Whoops. Now the debug message references a `$zone` variable that doesn't exist.

Sundypha: Right. I'll stop programming today. Already too late. I will review the whole thing in a few hours.

rmbolger: No worries. Take your time.

Sundypha:
Hi @rmbolger,

Did again another round of testing including timed out API calls to check debug and throw messages.
I figured now why my changes weren't persistent. VS Code opened the files at instdev.ps1 folder (\Documents\). First time writing any Powershell code.

Should I add another `$apiRoot` in the `Function Find-InfomaniakZone` for consistency?

rmbolger:
There should definitely be a local instance of `$apiRoot` within the function so you don't run into any unintended scoping issues. But whether you pass that instance as a parameter in the function or just hard code it doesn't really matter to me. The plugins are pretty small and having to find/replace an extra copy of the string if it ever changes is no big deal.

Sundypha:
I thought so exactly. I wasn't sure if Powershell has some kind of external scoping into functions called since it was working without but it felt awkward.

rmbolger:
Added one last fix that it looks like you had found originally and fixed in the Remove method but forgot to copy to the Add method.

rmbolger (approved):

Sundypha:
Hang on, I thought about this one. I'll try if the $RecordName can be used for adding the record instead of the short version.

Sundypha:
Yes, `source_idn` would work, but target_idn doesn't. The API request must have `target`.
We could change it to $RecordName, but I want to try with alias', too.
Will report shortly.

Sundypha:
With alias'ing the `source_idn` with `$RecordName` can't be used, so we stick with the `$recShort`.

rmbolger:
Thanks for the effort, @Sundypha. I'll try to get this out in a release soon.

PR #311, opened 2021-01-26
add support for acme-dns that uses http

AvrumFeldman:
Hi @rmbolger

Thanks so much for this great tool.

I had a niche need for http support with the acme-dns plugin. I hope it's not a problem to add it to the plugin.

I tested it using `Publish-Challenge`  and it passed.

Sample json to pass though

```
$reg = @{
    '_acme-challenge.www.example.com' = @(
        'ec6ec1f4-836e-462e-b577-b2f4e04d7291'
        'aeb00c77-852b-465e-9b27-984bc6cb12f5'
        'yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy'
        'ec6ec1f4-836e-462e-b577-b2f4e04d7291.acmedns.example.com'
    )
}

$pArgs = @{
    ACMEServer = 'acmedns.example.com'
    ACMERegistration = $reg
    HTTP = $true
}
```

Please excuse me if I made mistakes, I'm a rookie in git and open source in general (this is my very first PR).

### DIFF
--- a/Posh-ACME/Plugins/AcmeDns-Readme.md
+++ b/Posh-ACME/Plugins/AcmeDns-Readme.md
@@ -10,7 +10,7 @@ Once you have your instance running, it needs to be accessible via HTTPS to your
 
 ## Using the Plugin
 
-The only required parameter for the plugin is `ACMEServer` which is the hostname of the acme-dns instance you are using. If you have a setup that is different than the default acme-dns config, you might want to use the `ACMEUri` parameter instead tht allows you to provide the full URI to the server including protocol, port and path.
+The only required parameter for the plugin is `ACMEServer` which is the hostname of the acme-dns instance you are using. If you have a setup that is different than the default acme-dns config, if you customized your acme-dns deployment you might want to use the `ACMEUri` parameter instead that allows you to provide the full URI to the server including http/s, port and custom path.
 
 There is also an optional `ACMEAllowFrom` parameter which takes an array of strings with networks specified in CIDR notation. If included, these networks will be the only ones that can send TXT record updates to the server for the registrations that get created as part of the certificate request. In some environments, it may make more sense to control network access via standard firewalls.
 
@@ -27,7 +27,7 @@ _acme-challenge.example.com -> xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx.auth.acme-dn
 Press any key to continue.:
 ```
 
-For each CNAME in the list, you need to create the associated record on your DNS server before continuing. Assuming the records get created properly, the process should complete succesfully. Future renewals will complete without additional prompting as long as no new names are added to the cert.
+For each CNAME in the list, you need to create the associated record on your DNS server before continuing. Assuming the records get created properly, the process should complete successfully. Future renewals will complete without additional prompting as long as no new names are added to the cert.
 
 ## (Advanced) Pre-registration and CNAME creation
 
--- a/Posh-ACME/Plugins/AcmeDns-Readme.md
+++ b/Posh-ACME/Plugins/AcmeDns-Readme.md
@@ -10,7 +10,7 @@ Once you have your instance running, it needs to be accessible via HTTPS to your
 
 ## Using the Plugin
 
-The only required parameter for the plugin is `ACMEServer` which is the hostname of the acme-dns instance you are using. If you have a setup that is different than the default acme-dns config, if you customized your acme-dns deployment you might want to use the `ACMEUri` parameter instead that allows you to provide the full URI to the server including http/s, port and custom path.
+If your acme-dns instance is using the default configuration running with TLS on port 443, the only required parameter for the plugin is `ACMEServer` which is the hostname of your instance. If you are not using TLS, using an alternate port, or require additional uri path segments, you will need to specify the full URI using `ACMEUri` instead.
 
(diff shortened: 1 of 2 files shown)


rmbolger:
Hey @AvrumFeldman, thanks for the effort on the PR and even learning git for the purpose. That's great! I hadn't considered that people might not be running their acme-dns instance over TLS. But what if we go a step further with the change and make it possible to just specify your entire acme-dns root URL? That would allow for alternate ports as well or even routed sub-folders (e.g. `http://myserver.example.com:8080/acmedns`).

I'm picturing a second parameter set that uses something like `ACMEUrl` as the parameter name. So instead of the `$pArgs` like you have them, it would end up being like this:

```powershell
$pArgs = @{
    ACMEUrl= 'http://acmedns.example.com'
    ACMERegistration = $reg
}
```

It's a bit more extensive of a change to support it, but provides the most flexibility. I should have time to work on it this week if you don't.

AvrumFeldman:
@rmbolger Thanks for the encouragement.

I think your idea is great. Will work on it hopefully tonight.

But don't you think those changes should be separate PR's as they are different features?

(I did plan on submitting a separate PR for custom port, but after the first PR I couldn't recall what I planned on changing with the second PR :/ so thanks for reminding me ;)).

rmbolger:
The existing `ACMEServer` parameter should still work as-is because we don't want to break anyone who is currently using it. So the param attributes for it should change to specify a parameter set name, the new parameter should have a different parameter set name, and the CmdletBinding properties should specify the default as the one for ACMEServer. Basically this:

```powershell
[CmdletBinding(DefaultParameterSetName='Server')]
param(
    [Parameter(Mandatory,Position=0)]
    [string]$RecordName,
    [Parameter(Mandatory,Position=1)]
    [string]$TxtValue,
    [Parameter(ParameterSetName='Server',Mandatory,Position=2)]
    [string]$ACMEServer,
    [Parameter(ParameterSetName='Uri',Mandatory,Position=2)]
    [string]$ACMEUri,
    [string[]]$ACMEAllowFrom,
    [hashtable]$ACMERegistration,
    [Parameter(ValueFromRemainingArguments)]
    $ExtraParams
)
```

Anyone currently using `ACMEServer` will continue to work without needing to change anything. Anyone who needs the customization of `ACMEUri` can start using it. And the rest of the parameters apply to all parameter sets by default, so there's no change there.

Then within the function body, the first thing I'd do is basically convert ACMEServer to ACMEUri if that parameter set was used. Something like this:

```powershell
if ('Server' -eq $PSCmdlet.ParameterSetName) {
    $ACMEUri = "https://$ACMEServer"
}
```

Then it should just be a matter of tweaking all the `Invoke-RestMethod` calls to use `$ACMEUri` and updating the usage guide to talk about the ACMEUri option.

AvrumFeldman:
Added `ACMEUri`

Some notes: 
I don't see the reason we should need `DefaultParameterSetName='Server'`, any parameter without a setname is universal.
I think there has to be a stronger emphasis in the documentation that if using a hashtable for the parameters it has to be ordered.

rmbolger:
I like the intent of making ACMEUri an actual `[uri]` object, but (and I haven't tested this yet) I think it might break the custom param (de)serialization code in the module that exists now in 4.x unless I add a special case for it like I do for PSCredential and SecureString objects. That is to say, I think it might serialize properly to a JSON string, but then deserialize back to a `[string]` rather than a `[uri]` and then not work on renewals...though maybe PowerShell's auto-conversion stuff will be able to handle it? Not sure, would have to test. Though, I'd probably just make the parameter a string and call it a day.

Not having a DefaultParameterSetName may be ok in the current code base. But I've been bitten in the past with parameter binding issues when it's not there, so I'm overly cautious now.

AvrumFeldman:
> Not having a DefaultParameterSetName may be ok in the current code base. But I've been bitten in the past with parameter binding issues when it's not there, so I'm overly cautious now.

I removed it because I was debugging and issue and it seemed then that it was related. I will test with it back to see if we can add it.

Regardless just found that my code update is far from ready, as I found that I missed _(atleast)_ one location where the code needed to be updated to take into account the URI option. I already fixed that, but I don't want to push it yet to the PR branch as I need to test the whole plugin modification better and I just realized I was doing some testing wrong.

Hope to have some updates tomorrow.

AvrumFeldman:
This code was tested by me and worked. I tested both methods. Either by preregistering or by letting the plugin register. and did a renewal right away and it also worked. Disclaimer: I tested against Let's Encrypt Staging.

I'm just going to break apart the changes I made and why.
1. In the base module I added another parameter with a parameter set, so users can choose whichever one they want.
2. I think it makes sense we should only once evaluate whether it's a custom URI or host_address, so that's why I did it right in the beginning. 
3. I also think the helper function doesn't need to revalidate the URI provided as it should have already passed validation on the parent function. Even more, as we are already defining in the parent function the actual URI, the helper function should need need to know if it is a host_name or custom URI, all it should get is a URI that it uses, so I also changed that.
4. Last, I removed the line continuations. We can always use splatting if you don't like the long line.

rmbolger:
I tested the changes as well and decided to get rid of the uri param validation stuff just because it seemed overkill. If they screw up the URI, that's on them and `Invoke-RestMethod` will deal with throwing an error. Added the DefaultParameterSetName just to make sure the current parameter binding functionality remains unchanged. The URI parameter does end up getting serialized as a plain string in the JSON on disk. But it looks like PowerShell is smart enough to massage it back into a `[uri]` on renewals.

Added the splatting as you suggested. Tweaked the text in the guide a bit. All in all, I think this is good to go. Thanks for the effort!

AvrumFeldman:
No objections. Looks great. 

Was a pleasure working with you on this PR. Thank you.

PR #317, opened 2021-02-04
Handle externally-generated CSRs that have no attributes

methorpe:
Hi Ryan,

I've made a small patch to Get-CsrDetails so that it can handle CSRs that have no attributes, such as if there were no SANs in the CSR. Get-CsrDetails already handles CSRs that have a CN and non-SAN attributes but no SANs, so it was just the no-attributes-at-all case that it didn't like.

Of course it would be good if every program that generated CSRs automatically added the CN as a SAN, but for those that don't and which need to generate the private key themselves (eg if it's generated and stored in a hardware security module) this will allow Posh-ACME to be used to submit their CSRs.

Thanks!

Martin


rmbolger:
Ooh, nice! Thanks @methorpe. Would it be possible to add an example CSR with no attributes into the `Tests\TestFiles` folder? I'd like to add a test for this in the Pester tests, but I'm not sure how to generate one specifically like this.

methorpe:
Hi Ryan!

It's very good of you to respond so quickly -- thank you!

I've uploaded an example CSR and test case. I've never used Pester before so that was a good learning experience. The commit message has notes on how to regenerate the CSR if desired.

Many thanks, and best wishes,

Martin

rmbolger:
Perfect! Thanks a ton. You are awesome.

PR #325, opened 2021-03-13
Fix Submit-Renewal not sending all previous order params to New-PACertificate

juliansiebert:
Hi,
the keylength of the previous order does not get reused und renewal.

I don't know if there is something else missing, but this fixed it for me.

Best regards
Julian


rmbolger (approved):

rmbolger:
Fixes issue #326

PR #332, opened 2021-03-30
Note about renewal limits in tutorial

gomker:
I ran into an unexpected issue when I renewed a few certs in a root domain, I wasn't aware of the rate limit special condition, thought it might help to add to the tutorial.

### DIFF
--- a/Tutorial.md
+++ b/Tutorial.md
@@ -177,6 +177,8 @@ Submit-Renewal -AllAccounts
 # The -Force parameter works with these as well.
 ```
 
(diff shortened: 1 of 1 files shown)


rmbolger:
This is definitely one of the rate limits people miss even if they've skimmed the rate limits page. I like it. I'd probably put it up higher in that section though. Like after the paragraph that talks about using `-Force`.

And use the `> **_NOTE:_** ` markdown prefix to make it stand out more.

gomker:
Made the requested changes.

rmbolger:
Thanks for the update. Almost forgot about this. Made a small tweak myself. But otherwise, looks good.

PR #335, opened 2021-04-08
added a new DNS plugin for the All-Inkl.com KAS API

astaerk:
Hi, I need support for the KAS API of all-inkl.com. Therefore I created a new DNS plugin for it.

I already used the plugin on Win10 with PS7.1.3 and on Win2019 with PS5.1. So I'm positive it works with both versions of powershell.

During development I was unsure about the parameters regarding passwords and so on. Some plugins provide additional 'insecure' parameters. When I saw, that you deprecated some of them lately, I went with securestring parameters only. Is this fine?

Also this is the first time I've written that many lines of code in powershell. So don't hesitate to tell me what needs to be done differently :)

It seems as if one could create a test account: https://all-inkl.com/webhosting/test-account/
But I didn't test that yet. So I don't know if the test account supports API access. Also the site is in german.
According to the link the test account generally will be available within one business day and will be automatically disabled after 7 days.

### DIFF
--- a/Posh-ACME/Plugins/All-Inkl-Readme.md
+++ b/Posh-ACME/Plugins/All-Inkl-Readme.md
@@ -1,0 +1,23 @@
+# How To Use the All-Inkl KAS DNS Plugin
+
+This plugin works against the [All-Inkl](https://www.all-inkl.com/) KAS DNS provider. It is assumed that you have already setup an account and created the DNS zone(s) you will be working against.
+
+## Setup
+
+Using the All-Inkl KAS API only requires your accounts username and password.
+
+## Using the Plugin
+
+In any case you have to provide the username with the `KasUsername` parameter.\
+In addition you can provide the password as SecureString in plain text by using the `KasPwd` parameter or as sha1 hash by using the `KasPwdHash` parameter. In both cases this plugin only sends a sha1 hash of your password to the All-Inkl KAS API.
+
+Alternatively you can handle the authentication yourself and provide a valid session id by specifying the `KasSession` parameter as SecureString.
+
+```powershell
+$pwd = Read-Host "All-Inkl KAS Password" -AsSecureString
+$pArgs = @{
+    KasUsername = 'myusername'
+    KasPwd = $pwd
+}
+New-PACertificate example.com -Plugin All-Inkl -PluginArgs $pArgs
+```
--- a/Posh-ACME/Plugins/All-Inkl.ps1
+++ b/Posh-ACME/Plugins/All-Inkl.ps1
@@ -26,20 +26,15 @@ function Add-DnsTxt {
     $settingsObj = Get-KASDNSSettings $loginData $RecordName
 
     # remove zone from record name
-    $zoneIdx = $RecordName.LastIndexOf($settingsObj.zone) - 1 # -1 for also removing the .
-    if ($zoneIdx -ge 1) { # at least one character should be left after removing, so check against 1 and not 0
-        $recNameWithoutZone = $RecordName.Remove($zoneIdx)
-    } else {
-        $recNameWithoutZone = $RecordName
-    }
-    
+    $recNameWithoutZone = ($RecordName -ireplace [regex]::Escape($settingsObj.zone), [string]::Empty).TrimEnd('.')
+
     # search for existing DNS settings for the record
     $existingSettingsItem = Find-KASDNSSettingsItemInList $settingsObj.dnsSettings $recNameWithoutZone $TxtValue
     if ($existingSettingsItem) {
         Write-Verbose "Record $RecordName already contains $TxtValue. Nothing to do."
         return
     }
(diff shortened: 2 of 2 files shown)


rmbolger:
Hi @astaerk. Thanks for the submission! Looking forward to reviewing and testing it. But I probably won't be able to until maybe Sunday or Monday.

rmbolger:
Small update. I submitted a request for a test account this morning. But I haven't heard anything back yet.

rmbolger:
I'm not sure what's taking them so long, but just wanted to mention I haven't heard anything back on the test account yet. Normally at this point I'd just register a new domain with them. But the cheapest TLD they offer for a US user is 15 € and that's a little too steep for a throwaway domain.

astaerk:
Thanks for your effort. I'm not sure what happened there. I also requested a test account and I received a mail with the account data on the same day. Maybe it got lost in spam?
But anyway: I don't think it is possible to use a test account for testing the plugin.

The test account contained a subdomain only which is not visible in the DNS settings view.
I added a random second level domain to the settings because obviously nothing would go live until I transferred that domain to them.
That domain was visible in the DNS settings but it shows a banner that those settings cannot be changed with a test account.

I'll also have a look if I can come up with something.

rmbolger:
> Maybe it got lost in spam?

Indeed it had and I feel dumb for not checking. Thanks for the reminder though. I'll try playing around with the test access now that I have it.

> Btw. 15€? Really? Seems they don't want customers in the US :)

Yep. `.de` is available for 10€, but you have to have a non-PO Box address in Germany. The US registrars I've dealt with usually have deals for as little as $1 USD for the first year on the obscure TLDs. If you are able to setup a second account on a trial and send me the credentials (ryan-oss (at) xyto.cc), that would totally work. I shouldn't need it for more than a week.

rmbolger:
> That domain was visible in the DNS settings but it shows a banner that those settings cannot be changed with a test account.

Confirmed the same behavior on my test account. Let me know if you come up with an alternative.

rmbolger:
Quick question as I peruse the code. For the various things that use `$recNameWithoutZone`, have you tried a test where that value ends up the same as the root domain (where $RecordName *is* the domain instead of a sub like _acme-challenge)? The only way it would normally happen is using the DNSAlias feature. I think right now the value gets set to the domain. But some providers expect something like "@" or an empty string. The way I usually test is like this:

    $DebugPreference = 'Continue'
    Publish-Challenge fakedomain.com (Get-PAAccount) faketoken All-Inkl $pArgs -DnsAlias realkasdomain.com -Verbose
    Unpublish-Challenge fakedomain.com (Get-PAAccount) faketoken All-Inkl $pArgs -DnsAlias realkasdomain.com -Verbose

P.S. The code looks great, btw. Props for implementing a SOAP client without an actual SOAP library.

astaerk:
> If you are able to setup a second account on a trial and send me the credentials

I presumably check that tomorrow.

> For the various things that use `$recNameWithoutZone`, have you tried a test where that value ends up the same as the root domain

Actually I wasn't aware that could happen, so no. I will also try to check that tomorrow.

> P.S. The code looks great, btw. Props for implementing a SOAP client without an actual SOAP library.

Thanks!
Indeed I was checking out SOAP libs at first for a few minutes. But I quickly realized, that this would be overkill for those 3 messages. It's regular xml in the end.
That code is actually my second implementation. First I used System.Xml.Linq.XDocument. I had it all working with Publish-Challenge and Unpublish-Challenge for testing. But when it came down to really requesting a certificate it somehow crashed on unpublishing and I couldn't figure it out. Somehow the same inner methods were called many times for publish, but for unpublish I always got an error about some unknown properties on the .Net types. I couldn't figure it out so I tried to switch to the xml parsing I use now. Maybe the code works different after the 2 min timeout, who knows... :)

astaerk:
$RecordName being the domain itself indeed wasn't working. At one point in the code I made the assumption that it would always start with _acme-challenge. I pushed a new commit to fix that. Now I can successfully execute the code you mentioned to test that.

Regarding the account I sent you a mail. As mentioned in the mail I already tested it with this account and it worked for me.

Btw thank you very much for creating this project. I just replaced some old scripts (ACMEv1) of mine with posh-acme and it works like charm. Also posh-acme.deploy already came in handy. Great job!

rmbolger:
The root domain implementation was close, but not quite right. It was using the root domain as `$recNameWithoutZone` which in turn ended up creating a record like `example.com.example.com` instead of just `example.com`. I pushed a small fix for it, though my editor made it look bigger than it was due to a bunch of white space changes.

Thanks again for the test account. Feel free to delete it whenever you're ready.

astaerk:
Oh, of course. I was happy when I saw the root domain in the DNS settings. But obviously that created a subdomain. I didn't realize that at that moment. Sorry.

Thank you very much!

PR #336, opened 2021-04-11
Plugin to support easyname.com Hosting Provider

codemanat:
This plugins supports the easyname.com hosting provider for the ACME DNS-01 challenge. 

The code ist based on the easyname plugin for the dns-lexicon python library, which i fixed a while ago after easyname makes a complete website relaunch and the original plugin code stopped working. 

Since than i used the dns-lexicon easyname plugin without any issues.


rmbolger:
Hi @codemanat. Thanks for the plugin submission and excellent documentation. That's so weird that their REST API is so limited. I may take some extra time with this and see if I can make the web-scraping code less Windows-specific. But I won't be able to start until after I finish reviewing the other plugin PR.

rmbolger:
So...I started trying to remove the HTML parsing COM stuff so it'd be more cross platform and got a little carried away revamping things. No more COM dependencies. No more PSL dependency. It's all basically just regex. The credentials are now supplied via `EasynameCredential` rather than a separate email and password param. I think I also fixed a bug where the Add method would update the value of an existing record instead of just adding a new one. For LE challenges, it's normal to have multiple TXT records with the same FQDN and different values when doing a wildcard cert.

I think the biggest risk with the way the web scraping works now (and before actually) is that it will very likely break if the number of domains or number of records grows such that the web UI starts doing some sort of paging. (Ignoring that web scraping at all is inherently risky)

In any case, let me know how the update works for you.

codemanat:
@rmbolger 
Hi thank you for the adaptation. The code looks great now. But i think there must be a litte bug somewhere in the `Get-EasynameTxtRecord` or `Find-EasynameDomain` function. 

```
Get-EasynameTxtRecord : The argument cannot be bound to the ZoneID parameter because it is an empty string. 
In C: \Path\To \Location\Posh-ACME\Posh-ACME\Plugins\Easyname.ps1: 23 characters: 53 
+     if (Get-EasynameTxtRecord $ RecordName $ TxtValue $ zoneID) {
+                                                     ~~~~~~~ 
    + CategoryInfo: InvalidData: (:) [Get-EasynameTxtRecord], ParameterBindingValidationException
    +     FullyQualifiedErrorId: ParameterArgumentValidationErrorEmptyStringNotAllowed, Get-EasynameTxtRecord
```

Its very strange. With my test domain, the one i tested the prototype version against, there was not any error. After i testet against another domain, the above error where thrown. In my ongoing debug session, at some point, it suddenly started to work.

No issues while executing `Publish-Challenge`, `UnPublish-Challenge` anymore. So i changend to another domain name and the exact same issue occured. At first i suspected an easyname webserver problem but the allready "proccessed" domains where still working. 

It can't be the offset limit, becouse in the first version, this error was not present. BTW, according to there API docs, the offet limit starts at 50 domain entrys, which i didn't reach.  

The error did not always occur on a "fresh" domain sometimes it worked right away. 

The original Powershell Exception was thrown in german, so i translated it for you. Here is the original in case of some minor translation failures. :-) 

```
PS C:\...\Posh-ACME\Posh-ACME> Publish-Challenge example2.com -Account $acct -Plugin Easyname -Token faketoken -PluginArgs $pArgs -Verbose
AUSFÜHRLICH: Publishing challenge for Domain example2.com with Token faketoken using Plugin Easyname and DnsAlias
''.
Get-EasynameTxtRecord : Das Argument kann nicht an den Parameter "ZoneID" gebunden werden, da es sich um eine leere
Zeichenfolge handelt.
In C:\...\Posh-ACME\Posh-ACME\Plugins\Easyname.ps1:23 Zeichen:53
+     if (Get-EasynameTxtRecord $RecordName $TxtValue $zoneID) {
+                                                     ~~~~~~~
    + CategoryInfo          : InvalidData: (:) [Get-EasynameTxtRecord], ParameterBindingValidationException
    + FullyQualifiedErrorId : ParameterArgumentValidationErrorEmptyStringNotAllowed,Get-EasynameTxtRecord
```

At the moment, i didn't know what is going on here, but if i found the problem meyself, i will be posting an update.

Once again, thank you for your support! :-)

rmbolger:
Thanks for testing. This might be an issue with how I'm caching the results of `Find-EasynameDomain`. I recall a late change I made that I might not have fixed all the references for. I'll double check some things on my end too. When you test, it can be helpful to do it from a fresh PowerShell window to make sure no old variables are mucking things up.

rmbolger:
The other thing I just thought of is that `New-EasynameWebSession` tries to cache the session so it doesn't have to re-login for every action a cert might need. It probably easiest just to take this out, but I might try adding better error handling and only re-login if necessary.

rmbolger:
I ended up adding a wrapper around the `Invoke-WebRequest` calls that can detect an expired session by checking the response for the login page. It will then re-try the request after logging in again. I wasn't able to reproduce the issue you were having with the `$zoneID`, but I tweaked how those values are checked and I'm hoping the root cause was ultimately related to the session timeout.

Let me know how this version works for you.

codemanat:
@rmbolger 
Firstly thanks again for your effort. The Plugin Logic is so much better now. :-) 

Your ``New-EasynameWebSession`` wrapper  is great, i also thought about something like that. 
Yes, i also dumped the full PS Session while testing/debugging, just in case there are some $Variables lying around which i don't remember. :)

I just added a few minor changes but it seems to work now well. No error at all! I testet it about against 15 Domains or so. 

Changes in: ``Find-EasynameDomain``

There where duplicate ZoneID's in the ``$zoneID,$zoneName = Find-EasynameDomain $RecordName $EasynameCredential`` call.

**Fix**
````
#$id = ($domains | Where-Object { $zoneTest -eq $_.domain }).id
# Remove duplicates
$id = ($domains | Where-Object { $zoneTest -eq $_.domain } | Sort-Object -unique ).id
````

Changes in: ``Get-EasynameDomains``
For some reason, the `Invoke-WebRequest` Response in my Powershell Session is "German", even if the `Headers["Accept-Language"] = "en-US"` is added to the WebSession. So changed it for dealing with both circumstances.
 
**Fix**
````
$reDomains = [regex]'(?smi)<span class="domainname">(\S+)</span>.*?<a href="/de/domain/dns/index/domain/(\d+)">DNS|<a href="/en/domain/dns/index/domain/(\d+)">DNS'
````

Like i written before, it works great now. 

I just tested it against the staging environment, but that should not make any difference about how the DNS Records are created. :) 

````
New-PACertificate example.com -Plugin Easyname -PluginArgs $pArgs


Subject               NotAfter            KeyLength Thumbprint                               AllSANs
-------               --------            --------- ----------                               -------
CN=example.com        13.07.2021 23:59:59 2048      D66AECA5296F27DCE92F72F02BAA7B4F67813C8F {example.com}
````

Thank you. :)

**Comment by rmbolger** [comment]
Good catches on the duplicate IDs and URL differences between locales. You reminded me that I also didn't commit quite the correct version from my end. But even the version I meant to commit is probably still buggy due to language differences. I'm gonna work on it some more and see if I can make something that is not dependent on the locale.

**Comment by rmbolger** [comment]
Ok, I fixed the login page check in `Invoke-EasynameRequest` with the version I had meant to commit last time. I changed the regex things that were looking for URLs so they should be locale agnostic rather than only working for "en" or "de". And I changed the `Sort -Unique` in `Find-EasynameDomain` to `Select -First 1`.

If you can test this as well, that would be great. In particular, can you do a test that purposefully times out the session? What I've been doing is running a `Publish-Challenge` command, then leaving the window open doing nothing for 1 hour, then running the equivalent `Unpublish-Challenge` command. If you have `$DebugPreference = 'Continue'` on you should see it detect the login page and then re-login before trying again.

**Comment by codemanat** [comment]
@rmbolger 

Thank you, it seems to work now without any issue. The Re-Login is also working. 👍 

This time i needed a bit longer for testing, but its done now.

**Publish-Challenge**

```
Publish-Challenge example.com -Account $acct -Plugin Easyname -Token faketoken -PluginArgs $pArgs
DEBUG: Loading PAAccount list from disk
DEBUG: Calling Easyname plugin to add _acme-challenge.example.com TXT with value
gFihQhhycqdcsgnaTUSVNDgxSrEWXasGoefPffoZPSrqEg
DEBUG: GET https://my.easyname.com/en/domain/dns/index/domain/47110815
DEBUG: Post https://my.easyname.com/en/domain/dns/create/domain/47110815
DEBUG: Body:
{
    "id":  "",
    "name":  "_acme-challenge",
    "ttl":  "300", 
    "content":  "gFihQhhycqdcsgnaTUSVNDgxSrEWXasGoefPffoZPSrqEg",
    "type":  "TXT",
    "priority":  "0"
}
```
**Publish-Challenge (Nothing to do)**
```
DEBUG: Loading PAAccount list from disk                                                                                 
DEBUG: Calling Easyname plugin to add _acme-challenge.example.com TXT with value gFihQhhycqdcsgnaTUSVNDgxSrEWXasGoefPffoZPSrqEg
DEBUG: Searching for domains that match _acme-challenge.example.com                                               
DEBUG: GET https://my.easyname.com/en/login                                                                             
DEBUG: Attempting to login as admin@example.com                                                                     
DEBUG: POST https://my.easyname.com/en/authentication-api/login                                                         
DEBUG: GET https://my.easyname.com/domains/                                                                             
DEBUG: Checking _acme-challenge.example.com
DEBUG: Checking example.com
DEBUG: Found record match in domain example.com with id 47110815
DEBUG: GET https://my.easyname.com/en/domain/dns/index/domain/47110815
DEBUG: Found record with id 815123
DEBUG: Record _acme-challenge.example.com already contains gFihQhhycqdcsgnaTUSVNDgxSrEWXasGoefPffoZPSrqEg. Nothing to do.
```

**UnPublish-Challenge**
```
UnPublish-Challenge example.com -Account $acct -Plugin Easyname -Token faketoken -PluginArgs $pArgs
DEBUG: Loading PAAccount list from disk
DEBUG: Calling Easyname plugin to remove _acme-challenge.example.com TXT with value
gFihQhhycqdcsgnaTUSVNDgxSrEWXasGoefPffoZPSrqEg
DEBUG: GET https://my.easyname.com/en/domain/dns/index/domain/47110815
DEBUG: Response contains login page, attempting to re-login and try again.
DEBUG: GET https://my.easyname.com/en/login
DEBUG: Attempting to login as admin@example.com
DEBUG: POST https://my.easyname.com/en/authentication-api/login
DEBUG: Found record with id 815123
DEBUG: Post https://my.easyname.com/en/domain/dns/delete/domain/47110815/id/815123/confirm/1
```

rmbolger:
Excellent, I think this is ready to merge.

PR #338, opened 2021-04-21
Add ukfast dns plugin

Overglazed:
Hi 

This plugins supports the UKFast SafeDNS hosting provider for the ACME DNS-01 challenge.

I've followed the structure of the existing plugins as much as possible for easy maintenance.

Thanks :)

### DIFF
--- a/Posh-ACME/Plugins/UKFast-Readme.md
+++ b/Posh-ACME/Plugins/UKFast-Readme.md
@@ -4,7 +4,7 @@ This plugin works against the [UKFast](https://ukfast.co.uk/) provider. It is as
 
 ## Setup
 
-Authentication against the UKFast SafeDNS API requires an `API Application Key`, which can be obtained from the [MyUKFast](https://my.ukfast.co.uk/applications/index.php) portal. Additionally the key must be given 'READ/WRITE' permission for the SafeDNS service. 
+Authentication against the UKFast SafeDNS API requires an `API Application Key`, which can be obtained from the [API Applications](https://my.ukfast.co.uk/applications/index.php) portal. Additionally the key must be given `READ/WRITE` permission for the SafeDNS service.
 
 For further documentation on working with API keys, please visit the [UKFast Developer Centre](https://developers.ukfast.io/getting-started#registering-applications).
 
--- a/Posh-ACME/Plugins/UKFast-Readme.md
+++ b/Posh-ACME/Plugins/UKFast-Readme.md
@@ -10,23 +10,9 @@ For further documentation on working with API keys, please visit the [UKFast Dev
 
 ## Using the Plugin
 
-The API key will be used with either the `ApiKey` or `ApiKeyClearText` parameter. 
-- `ApiKey` will use SecureString. 
-- `ApiKeyClearText` will use a standard cleartext string.
-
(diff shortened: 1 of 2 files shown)


rmbolger:
Thanks for the submission, @Overglazed. I started trying to create an account to test with, but I couldn't complete the registration for some reason. So I opened a ticket with the accounts team, but haven't heard back yet.

Two small things I noticed in the code so far.

Instead of using the literal `-UseBasicParsing` parameter on the rest calls, use `@script:UseBasic` instead which will prevent errors if/when future versions of PowerShell remove that deprecated parameter from the web cmdlets. The module creates the variable when it's imported.

As of 4.x, you don't need to include the insecure/clear text version of secure params anymore and I've started deprecating them on older plugins. Also, the param names should be plugin specific. So something like `$UKFastApiKey` rather than just `$ApiKey` to prevent naming collisions with other plugins.

Overglazed:
Thanks for the comments @rmbolger. 

I've reverted back to use `@script:UseBasic`, as I had changed that to do some testing of the individual cmdlets then forgot to switch it back! Apologies.

I've also updated the parameter name to `$UKFastApiKey`, and removed the insecure/clear text option. 

Please let me know if you have any problems with testing. 

If you don't hear back about an account soon, I'll see if I can get something in place for testing.

Overglazed:
Hi @rmbolger.  Were you able to register for an account successfully in the end? 

I can speak to the team and find out what happened to your registration email if you can let me know the username you signed up with. :)

Thanks,
Lee

rmbolger:
Unfortunately, not. I never got a response from the team aside from the initial auto-response and I've been too busy to follow up on it this week. The username I tried to sign up with is "rmbolger".

Overglazed:
I have spoken to our accounts team and they have managed to tracked down and enable the `rmbolger` account. Seems like registration went through but the account activation didn't complete. We're looking into what caused that.

You should now be able to log in to the https://my.ukfast.co.uk portal.  If you're not sure of the password, you can use the password reset link on the login page. 

Let me know if you have any further issues :)

rmbolger:
Thanks! It seems like the password I created during setup is working, but something is still wonky with the account and after login, I get redirected to what looks like a signup related URL that results in a 404. `https://www.ukfast.co.uk/myukfast-signup/(guid-like-string).html?msg=6&url=%2Fdashboard.php&err=401`

The password reset functionality works. But I get the same login redirect after using the new password.

rmbolger:
This is pretty comical. The next time I logged in, I got a Site Maintenance message. Tried again a few minutes later and now it's trying to send a 2FA code to...somewhere. All the screen shows is `***** *** *` as the destination. But I don't recall giving my phone number during registration...so I'm not sure where it might be going. But it didn't end up getting to my phone or email.

No worries on all this, by the way. It's kind of funny. Is it worth having them just delete the account and me trying again from scratch?

Overglazed:
Well thats annoying!  Sorry for this. The account provisioning process may not have completed properly after the account needed that manual intervention.  I'll raise this with the team in charge of that process and see if we can remove this account and start from scratch, as you suggest.

It's a public holiday in the UK today, so I'll provide an update here tomorrow once I speak to the team.

Overglazed:
Hi @rmbolger. I was on leave most of last week, apologies for the wait.  The team in charge of registration tells me that the issues  with your account have been resolved.  2FA has been disabled to allow access, since this was never set up during registration. 

You should be able to login with your existing credentials.  

Thanks,
Lee

PR #340, opened 2021-04-29
Company is called Simply.com, not Simply :)

tomsommer:


rmbolger:
Thanks, @tomsommer. Always appreciative of corrections.

rmbolger:
P.S. While I have you here, Tom. Does Simply.com have any plans for a v2 API that does not require putting the API Key directly in the endpoint URL? Having it in the URL means the value is likely getting leaked all over log files for things that use it.

tomsommer:
@rmbolger We recognize the API is not optimal due to being built quickly and in another decade. At some point we will rewamp it :)

rmbolger:
Totally understandable. Thanks.

PR #341, opened 2021-04-29
Added support for IDN domains in the Simply.com plugin

Norskov:


Norskov:
@tomsommer FYI. Based on my ticket earlier today.

rmbolger:
Hey @Norskov, thanks for the submission. Can you verify my understanding for the context of the change? The Simply.com API does not support referencing the zone/domain in the endpoint URLs using the punycode version of the IDN name. This change fixes the plugin by converting detected punycode domain names into the actual unicode string to then pass to the various Invoke-RestMethod calls.

tomsommer:
The correct method is doing a lookup in `/my/products/` and using the return [object] string matching the domain, in the following queries, since the object-reference in theory can be anything.

rmbolger:
That's an excellent clarification. Thanks. Do you think it's worth changing the way the domain is found in the plugin so the name assumption doesn't break things in the future? Also, does `/my/products` have any max result limits or paging to be aware of for people with really large accounts?

rmbolger:
@tomsommer Might there be a way to waive the 7 EUR dns setup fee on a new IDN domain purchase I'm trying to make so I can better test this? Feel free to contact me at ryan-oss (at) xyto.cc if you'd rather speak offline.

rmbolger:
@tomsommer and @Norskov. I'm going to merge the changes as they exist. But I'd really like to do some further testing on an IDN domain and possibly revamp how the zone is found to better align with the intent of the API like Tom mentioned. It looks like I can purchase a .se domain for 0.90 EUR, but there's an additional 7 EUR DNS setup fee that's a bit steep for my hobbyist budget. Any way you guys could sponsor me for the difference or send me a coupon code to waive the setup fee? The CTO has to have some pull, right?

Norskov:
@rmbolger Thank you for having a look at it. I've sponsored you for $15 which should cover the costs.

rmbolger:
I'm not sure what happened. I saw the notification come through, but it looks like it got cancelled.

Norskov:
> I'm not sure what happened. I saw the notification come through, but it looks like it got cancelled.

Yeah, sorry about that. I've tried various things, and I'm not able to sponsor. I tried recovering my PayPal account, but that didn't help either. I keep getting the error "Your most recent billing attempt failed. We will try again on February 18, 2017.". Is there any other way I can send you the $15? @rmbolger

rmbolger:
Before GitHub sponsors existed, I was using PayPal directly (https://paypal.me/rmbolger). But if they're ultimately the problem, that probably won't work. It doesn't look like Venmo works internationally which is the other service I'm currently registered with. I could try signing up for [Wise](https://wise.com/)?

Or is there a way to generate a coupon code for me that just discounts the order by 7 EUR or credit me after the fact?

rmbolger:
I got a notification that a $15 one-time sponsor came through for you yesterday. The activity history doesn't show it being cancelled like the previous one. Did the PayPal issue get sorted?

tomsommer:
Version 2 of the API now uses HTTP-Basic auth to hide the login from logs

PR #343, opened 2021-05-06
Add dedicated Revocation function that supports additional use-cases

methorpe:
Hi Ryan,

I've added a public function to revoke a certificate -- this merely calls out to the CA to request that the certificate be revoked, it doesn't update any of the local state on the machine. I think I've covered all the bases but please feel free to modify the code as you see fit.

My company (Venafi) has been developing a connector to allow our machine identity management software to connect to ACME CAs, and for what we wanted to do, Posh-ACME looked to be the best ACME library -- I hope you'll consider my patches as a way of contributing a little back!

Thanks for a super library.

Best wishes,

Martin


rmbolger:
Hi @methorpe. Thanks for the submission! Interestingly, I had actually been thinking about adding some additional functionality around revocation recently. So this is good timing. However, I think there might be some commits missing in the PR. The function as I can see it currently ends with `$cert.GetEncoded()` and never actually talks to the CA. Do you have some local commits that didn't get pushed to the Github fork?

That said, the shape of the function looks like it might be duplicating what you can already do using `Set-PAOrder`. For example:

```powershell
Revoke-PACertificate example.com
```

looks functionally equivalent to 

```powershell
Set-PAOrder example.com -RevokeCert -Force
```

methorpe:
Hi Ryan,

*brown bag moment*

Haha, yes, I know why :) I’d started making the modifications on the machine I’d cloned the repo on, and then ended up finishing the job on the machine that was actually Posh-ACME. And then, of course, forgot to push the changes back to the repo. Thank you for being more on the ball than I was!

And, even more embarrassingly, I had not noticed the -RevokeCert argument to Set-PAOrder. I will give that a try and report back, but it looks likely that you won’t need this pull request after all.

Thank you, and apologies!

Best wishes,

Martin

rmbolger:
No worries. And to be fair, the existing revocation functionality isn't exactly the easiest thing to find. And given all the [different ways that exist](https://letsencrypt.org/docs/revoking/) to revoke a cert, pulling all that functionality out into a dedicated function is probably a good idea long term.

rmbolger:
Thanks, @methorpe. Being able to pass a cert file definitely makes sense for some of the revocation options I had been thinking about. If you don't mind, I'm gonna use your PR as a sort of feature-branch to hack on while I flesh out the ideas I have in my head.

Here are some usage variations that I'm hoping to incorporate. All of these will likely prompt for confirmation unless a `-Force` switch is passed. They'll all also have the optional Reason param, though I may change it to a string based enum rather than the int value it currently is.

- `Revoke-PACertificate` *(no params)*
  - Revokes the cert associated with the current order (if one exists)
- `Revoke-PACertificate example.com`
- `Get-PAOrder example.com | Revoke-PACertificate`
- `Get-PACertificate example.com | Revoke-PACertificate`
  - Revokes the cert associated with the order for `example.com` (if one exists)
- `Revoke-PACertificate -CertFile path\to\cert.cer`
  - Revokes the cert associated with the specified cert file. If the cert wasn't acquired with the current ACME account, this will likely throw an error about needing to validate the names in the cert on this account first. It would be nice if it could have the ability to do that validation as part of the command, but that would involve duplicating a lot of functionality that already exists in  things like `New-PACertificate`.
- `Revoke-PACertificate -CertFile path\to\cert.cer -KeyFile path\to\cert.key`
  - Revokes the cert associated with the specified cert file and private key. This code path should succeed even if the current ACME account didn't acquire the original cert because we can sign the request with the cert's private key.

rmbolger:
Ok, I think I've got all of the basics I was hoping to add implemented.

The function's ConfirmImpact is set to High which means it will prompt for confirmation unless you pass `-Confirm:$false` or `-Force` or the calling scope's `$ConfirmPreference` variable is set to None.

You can still revoke arbitrary cert files even if they weren't created with Posh-ACME as long as they're PEM formatted. If they were created with a different ACME account, you'll either need to pass the associated private key using `-KeyFile` or create a dummy order that contains the same set of names as the cert you're trying to revoke and validate the challenges for them first. I should probably write up a wiki guide on this at some point.

The MainDomain param provides tab-completion like the other functions with a MainDomain param.

I added all of the reason codes to the enum as referenced in RFC 5280 because the ACME spec doesn't place any limitations on what's allowed and leaves it up to each CA to choose what they allow. So if you try to send a reason that Let's Encrypt (or whoever) doesn't support, you'll just get an ACME error that it's not supported and a list of ones that CA does support. The reason codes also tab-complete.

@methorpe, thanks again for the push to get this implemented. If Venafi is seriously going to start relying on the module, I'd be grateful if you could try and get them to [sponsor](https://github.com/sponsors/rmbolger) my work as well. 😁

PR #347, opened 2021-05-12
Add non-interactive switch for Manual plugin

hhhuut:
Hi there,

I recently ran into trouble trying to automate my certificate management. My provider unfortunately doesn't offer an API to create/modify DNS records, so I was stuck with the `Manual` provider, however my automation pipeline always breaks when the user is prompted to press Enter to continue.

So I thought the easiest way around that is to add a new provider that effectively does exactly the same as the `Manual` one (printing the records to add to the console) except immediately starting with the wait period of `DnsSleep` instead of prompting for input.

Best regards,
hhhuut


rmbolger:
Hi @hhhuut. How about instead of a whole new plugin, just an optional switch for the existing Manual plugin that disables the prompt? Perhaps something like `[switch]$ManualNonInteractive`?

hhhuut:
Thanks for your feedback.
I thought about that aswell but wasn't sure what the preferred approach would be, however I removed the additional provider now and added a switch instead, just like you suggested.

rmbolger:
I made some minor changes to better align with how plugin args normally work.

But before I merge this, I'm still a little confused about how this actually helps your workflow. Like, if your automation workflow can now proceed without being interrupted, wouldn't it mean you never see the TXT records that need to be created for each renewal? Or is this more of a manual renewal that you're actively watching? And when the TXT record details scroll by you basically race the DnsSleep timer to add the records by hand to your DNS server?

hhhuut:
> I made some minor changes to better align with how plugin args normally work
Cool, I apologize if my code didn't meet your standards, PS scripting is not really my strong suit.

> wouldn't it mean you never see the TXT records that need to be created for each renewal?
I'm doing this for a small software start-up and we're using Azure Pipelines to automate most of our workflows. It doesn't allow user-input/prompts, however it captures and displays console output, so the TXT records that are printed are visible.

> And when the TXT record details scroll by you basically race the DnsSleep timer to add the records by hand to your DNS server?
Yep. It's not a perfect solution, I know believe me, however our DNS provider doesn't offer an API to access records, so this level of (partial) automation is better than none.

rmbolger:
Gotcha. No worries. If you haven't looked into it yet, [acme-dns](https://github.com/joohoi/acme-dns) is a common solution to not having API access to the DNS servers. There are also a number of free DNS providers out there with API access you could potentially migrate to that also have Posh-ACME plugins. Cloudflare and DigitalOcean are two popular ones.

hhhuut:
Thanks man, appreciate it.

Yeah, I've thought about migrating aswell but that'll be a bigger task ^^

Have a good one.

PR #351, opened 2021-05-22
Aurora-Plugin-v2021.0522.1930

j81blog:
Initial version of the DNS Aurora-Plugin (https://pcextreme.nl/)

### DIFF
--- a/Posh-ACME/Plugins/Aurora-Readme.md
+++ b/Posh-ACME/Plugins/Aurora-Readme.md
@@ -13,10 +13,10 @@ With your API key and secret, you'll need to pass them with the `Credential` par
 ```powershell
 # Prompt for the credential
 $auroraCredential = Get-Credential -Message "Aurora Username:Key / Password:Secret"
-$auroraParams = @{ ApiAurora='api.auroradns.eu'; CredentialAurora=$auroraCredential }
+$auroraParams = @{ AuroraApi='api.auroradns.eu'; AuroraCredential=$auroraCredential }
 
 # Or entering the key and secret
-$auroraParams = @{ ApiAurora='api.auroradns.eu'; CredentialAurora=$((New-Object PSCredential 'KEYKEYKEY',$(ConvertTo-SecureString -String 'SECRETSECRETSECRET' -AsPlainText -Force))) }
+$auroraParams = @{ AuroraApi='api.auroradns.eu'; AuroraCredential=$((New-Object PSCredential 'KEYKEYKEY',$(ConvertTo-SecureString -String 'SECRETSECRETSECRET' -AsPlainText -Force))) }
 
 # Request the cert
 New-PACertificate example.com -Plugin Aurora -PluginArgs $auroraParams
--- a/Posh-ACME/Plugins/Aurora-Readme.md
+++ b/Posh-ACME/Plugins/Aurora-Readme.md
@@ -8,15 +8,15 @@ If you haven't done it already, you need to generate API Credentials for your ac
 
 ## Using the Plugin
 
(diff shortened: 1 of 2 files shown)


rmbolger:
Hi @j81blog. Thanks for the submission. Out of curiosity, why is the plugin referred to as Aurora instead of something like PCExtreme?

It may be a couple days before I can dig into this for testing. But there are a couple things you can probably deal with in the mean time.

- The Secure and Insecure parameter set differentiation isn't technically necessary anymore. So you can simplify things and remove the Insecure one if you want. 
- Because you can use multiple different plugins for a single certificate, the parameter names for the plugin need to be unique so they don't potentially conflict with another plugin. So instead of just `$Credential` and `$Api`, make them  something like `$AuroraCredential` and `$AuroraApi`.

Another curiosity question I had was about the double type definition before the `$Credential` param. I haven't seen that before and usually only use a single `[pscredential]` type identifier. What is the purpose of having both?

       [System.Management.Automation.PSCredential]
       [System.Management.Automation.Credential()]

j81blog:
Hi Ryan,

The service they provide is called Aurora DNS, used this as soure:
[Aurora DNS API Docs](https://libcloud.readthedocs.io/en/latest/dns/drivers/auroradns.html#api-docs)

Per request removed the insecure parameters.

About the following:
[System.Management.Automation.PSCredential]
[System.Management.Automation.Credential()]

this combination will prompt you to enter credentials if it's required and you don't have any credentials specified.
https://docs.microsoft.com/en-us/powershell/scripting/learn/deep-dives/add-credentials-to-powershell-functions?view=powershell-7.1
"Using (Get-Credential) seems cumbersome. Normally, when you use the Credential parameter with only a username, the cmdlet automatically prompts for the password. The [System.Management.Automation.Credential()] attribute enables this behavior."
&
https://duffney.io/addcredentialstopowershellfunctions/

rmbolger:
Sorry for the delay. It took longer than I was expecting to get the test domain registered. The basic functionality is working great, nice work! There's one edge case bug I ran into that could use fixing though. It's basically when the `$RecordName` parameter matches the zone name which can happen when someone uses a challenge alias against the apex of the zone. It looks like `$recordPath` ends up as an empty string and some of the helper functions don't currently allow that.

You can test this with Publish-Challenge and the `-DnsAlias` parameter where "realdomain.com" is your actual hosted zone and "fakedomain.com" is whatever. The fake domain never gets sent to the plugin, so it doesn't really matter what it is.

```powershell
$DebugPreference = 'Continue'
Publish-Challenge fakedomain.com (Get-PAAccount) faketoken Aurora $pArgs -DnsAlias realdomain.com -Verbose
Unpublish-Challenge fakedomain.com (Get-PAAccount) faketoken Aurora $pArgs -DnsAlias realdomain.com -Verbose
```

Another thing I was curious about is the code revolving around the `-UseBasicParsing` flag. It looks like you're trying to make the plugin capable of being dot sourced and used without having to deal with errors because `$script:UseBasic` doesn't exist outside the module scope. But in the process, I think you might be accidentally breaking the future-proofing aspect of `$script:UseBasic`.

Because the `-UseBasicParsing` parameter is deprecated, it might not exist in future versions of PowerShell. In that theoretical future version, `$script:UseBasic` should be an empty hashtable which would end up doing nothing when splatted against the web cmdlets. But all of the calls in the plugin are explicitly adding `-UseBasicParsing:$UseBasicParsing` which means they'd break in that future version.

If you need the plugin to be easily usable outside of the module, I'd probably change things so you build your own instance of the UseBasic hashtable based on the same code the module uses where hashtable is empty unless the parameter actually exists.

```powershell
$UseBasic = @{}
if ('UseBasicParsing' -in (Get-Command Invoke-WebRequest).Parameters.Keys) {
    $UseBasic.UseBasicParsing = $true
}
```

j81blog:
Thank you, made the necessary changes as requested.

rmbolger (approved):
I made a few tweaks to the readme. But everything else looks great. Thanks again, @j81blog.

PR #353, opened 2021-05-30
Change Aurora DNS

j81blog:
Made some small adjustments. Had some issues with previous version with PowerShell Core.
Now tested on PowerShell 7 (Windows).


rmbolger:
That's odd. I had been testing on PS 5.1 and 7.1.3 without a problem on my Win10 20H2 machine. Usually, I run into version compatibility errors like this with old versions rather than new versions. In any case, everything still works for me.

j81blog:
Maybe the issue didn't occur in situations with Posh-ACME. Great it is added!
Thank you again for al your great work!

PR #363, opened 2021-07-26
Correct Windows PowerShell reference in first line.

homotechsual:
This changes the reference to Windows PowerShell in the first line to "PowerShell" with a link to the platform requirements as PowerShell 7 (and PSCore) are now supported as well.


rmbolger:
Thanks for the PR. This is definitely one of those things that I never remember to re-validate over time.

PR #373, opened 2021-08-21
Added New-PAAuthorization

rmbolger:
Some ACME servers may support [pre-authorization](https://datatracker.ietf.org/doc/html/rfc8555#section-7.4.1) which allows users to create and validate authorization objects outside of the normal order lifecycle.

`New-PAAuthorization` enables users to request new authorization objects for a given set of identifiers. If authorization objects already exist for a given identifier, the ACME server will return the existing object.

Currently, the only public ACME CA that seems to support pre-authorization is BuyPass.

### DIFF
--- a/Posh-ACME/Posh-ACME.psd1
+++ b/Posh-ACME/Posh-ACME.psd1
@@ -36,6 +36,7 @@ FunctionsToExport = @(
     'New-PAAccount'
     'New-PACertificate'
     'New-PAOrder'
+    'New-PreAuthorization'
     'Publish-Challenge'
     'Remove-PAAccount'
     'Remove-PAOrder'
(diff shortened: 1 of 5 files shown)


indy-singh:
I hate bumping old threads; but just wanted to say thanks for official supporting pre-auth. One of the very few clients to do so and makes it super easy:-

```pwsh
Install-Module -Name Posh-ACME -Scope CurrentUser;
Set-PAServer -DirectoryUrl "https://acme.enterprise.sectigo.com";
New-PAAccount -ExtAcctKID $kid -ExtAcctHMACKey $hmac -Contact $email -AcceptTOS;
New-PACertificate @(Get-Content ..\domains.txt) -AcceptTOS -Contact $email;
```
That's all it took. Amazing experience compared to other clients (looking at you certbot).

Many thanks,
Indy

PR #386, opened 2021-09-21
Use aws credentials from environment variables if no other options are set

plenti-chris-fowles:
If no other credentials are set we'll attempt to use environment variables for AWS credentials when using Route53.

We'll attempt to use AWS_SESSION_TOKEN only if it's also set.


plenti-chris-fowles:
this wasn't intended to be pushed upstream yet 🤦

rmbolger:
No worries. I like the idea anyway. Looking forward to the implementation. 👍

Issue #381, opened 2021-09-14
Plugin for Core Networks


dwydler:
Hello @rmbolger,
i try to write a Plugin for the german webhoster Core Networks. I have two questions about this:

1. According to the api documentation (https://beta.api.core-networks.de/doc/#functon_dnszones_commit) one or more changes must be commited. Only then are changes written from the database to the DNS zone. Is it possible that the function Save-DnsTXT get also the parameter `$RecordName` (optional)? I don't want to write my own function for it.

2. [UseBasicParsing](https://github.com/rmbolger/Posh-ACME/blob/c7ca310bf351636c3fd443e63cbfcac12e5d8568/docs/Plugins/Plugin-Development-Guide.md#usebasicparsing): Is that still recommended? If it's like that should i use `@script:UseBasic` or `$script:UseBasic`. In my tests (no matter which parameter of the two) the query no longer works. Here a query example of my tests:
 `Invoke-RestMethod -Method Post -Headers $headers -Body $JsonBody -ContentType "application/json" -Uri "$CoreNetworksApiRoot/dnszones/$CoreNetworkDnsZone/records/" @script:UseBasic`
What am I doing wrong?

rmbolger:
Hey @dwydler, that's great! Happy to help answer your questions. Feel free to post additional ones as you have them.

Regarding the Save function, no. Since it's only called once for all change in an order, the only parameters that are sent to it are the ones passed by the user via `-PluginArgs`. However, if you need some data like `$RecordName`, you should be able to save it during the Add/Remove calls as a [script-scoped](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_scopes#powershell-scopes) variable which will still exist when Save is called. There's actually a decent example of this in the [Manual](https://github.com/rmbolger/Posh-ACME/blob/main/Posh-ACME/Plugins/Manual.ps1) plugin where it collects the records from each Add/Remove call and then displays them to the console during the Save call.

UseBasicParsing is still definitely recommended. The variable normally only exists in the module's scope. So if you're testing that `Invoke-RestMethod` call outside the scope of a running plugin, you'd have to create your own copy like this:

```powershell
$script:UseBasic = @{UseBasicParsing=$true}
```

[...]

PR #388, opened 2021-09-24
Added plugin for core-networks.de

dwydler:
Hi @rmbolger,
as discussed (#381) here the Pull Request for my Plugin.
Your review is appreciated. I'm new in this area of development.

Regards,
Daniel

### DIFF
--- a/Posh-ACME/Plugins/CoreNetworks-Readme.md
+++ b/Posh-ACME/Plugins/CoreNetworks-Readme.md
@@ -1,0 +1,3 @@
+This guide has moved here:
+
+https://poshac.me/docs/v4/Plugins/CoreNetworks/
--- a/Posh-ACME/Plugins/CoreNetworks-Readme.md
+++ b/Posh-ACME/Plugins/CoreNetworks-Readme.md
@@ -1,3 +1,0 @@
-This guide has moved here:
-
-https://poshac.me/docs/v4/Plugins/CoreNetworks/
--- a/Posh-ACME/Plugins/CoreNetworks.ps1
+++ b/Posh-ACME/Plugins/CoreNetworks.ps1
@@ -200,7 +200,7 @@ function Save-DnsTxt {
     $script:CoreNetworksZones | Sort-Object -Unique | ForEach-Object {
 
         $queryParams = @{
-            Uri = "$ApiRootUrl/dnszones/$_/records/commit"
+            Uri = "$CoreNetworksApiRoot/dnszones/$_/records/commit"
             Method = 'POST'
             Headers = $headers
             ContentType = 'application/json'
--- a/Posh-ACME/Plugins/CoreNetworks.ps1
+++ b/Posh-ACME/Plugins/CoreNetworks.ps1
@@ -23,11 +23,6 @@ function Add-DnsTxt {
     }
     Write-Debug $headers
 
-    ### Convert an IDN name to punycode
-    $idn = [System.Globalization.IdnMapping]::new()
-    $RecordName = $idn.GetAscii($RecordName)
-    Write-Debug $RecordName
-
     ### Search und find the dns zone of the (sub)domain  (for example: example.com).
     $CoreNetworkDnsZone = $(Find-CoreNetworksDnsZones $CoreNetworksApiRoot $headers $RecordName)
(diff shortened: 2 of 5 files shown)


rmbolger:
Looking great, @dwydler. I did a bit of code cleanup for better readability and debug logging as well as some help/guide tweaks. Can you make sure I didn't break anything? I don't have an account with Core Networks at the moment, so I'm doing this blind.

I'd also like to take a stab at moving the Commit functionality into the Save-DnsTxt function.

dwydler:
> Looking great, @dwydler. I did a bit of code cleanup for better readability and debug logging as well as some help/guide tweaks.

Thanks for the improvements. 

> Can you make sure I didn't break anything?

Sure. I just had to change the type of a variable (see last commit). After the change the request of ssl certificates runs successfully.

> I'd also like to take a stab at moving the Commit functionality into the Save-DnsTxt function.
Like as discussed in the issue #381 i'm not a fan of global script variables.

rmbolger:
Good catch on the types. I tweaked them to be a bit more specific though. Can you do me a favor and run a publish/unpublish test that mimics a DNS Alias being used like this? The `fake.tld` can be anything and doesn't have to exist. The `realzone.com` should be a zone you actually have on Core Networks but make sure it's the apex of the zone. If you already tested this, great! Just wanted to double check.

```powershell
$pArgs = @{ CoreNetworksCred = (Get-Credential) }
Publish-Challenge fake.tld (Get-PAAccount) faketoken CoreNetworks $pArgs -DnsAlias realzone.com
Unpublish-Challenge fake.tld (Get-PAAccount) faketoken CoreNetworks $pArgs -DnsAlias realzone.com
```

I know you had mentioned your dislike of global variables, but I may have to insist. The API documentation explicitly talks about not being suitable for high-frequency processing and I'd rather not abuse their service by calling commit for every single record added or removed. In the worst case, a certificate with 100 names would end up being over 400 calls between the adds, removes, and commits. But you cut that almost in half if you only commit per zone (depending on the number of unique zones).

I'd be willing to compromise and add a switch parameter like `CoreNetworksImmediateCommit` that would let users choose to commit during the Add call instead of waiting for the Save call. But I want the default functionality to be respectful of the provider.

dwydler:
> Can you do me a favor and run a publish/unpublish test that mimics a DNS Alias being used like this?

Sure. Can you tell me what value the variable faketoken shoud be?

dwydler:
> I know you had mentioned your dislike of global variables, but I may have to insist

Okay. With the last commit I adjusted the behavior accordingly.

rmbolger:
> Sure. Can you tell me what value the variable faketoken shoud be?

The faketoken can literally be `faketoken` or really any string. I usually just use a single character like `a` because it's less to type.

rmbolger:
I tweaked the Save stuff to only commit once per unique zone name and reverted some of the variables you had converted to script variables. Essentially, we only need a single script variable to keep track of the zones we've added records to. I think I need to document how the plugin Save flow works better. Notably, it gets passed the same set of plugin args sent by the user as Add/Remove do. The only thing it doesn't get is `$RecordName` and `$TxtValue`.

I'm also going to ponder about a better way for Add/Remove to pass constructed data needed for Save. But realistically, that's not something that can happen until 5.x.

In any case, give this last version a test and if everything works, I think it's good to merge.

dwydler:
> The faketoken can literally be faketoken or really any string. I usually just use a single character like a because it's less to type. The TxtValue that actually gets saved to DNS will still look like gibberish because it's a base64 encoded hash of the key auth value.

The tests have been carried out again. It runs without any error. 
Below are my test commands. To make sure i done it right.
```
Publish-Challenge test.de (Get-PAAccount) a CoreNetworks $pArgs -DnsAlias "acme.domain.de"
Unpublish-Challenge test.de (Get-PAAccount) a CoreNetworks $pArgs -DnsAlias "acme.domain.de"
```

> I tweaked the Save stuff to only commit once per unique zone name and reverted some of the variables you had converted to script variables.

I review you changes and now it make sense. Thanks for the improvements.

> In any case, give this last version a test and if everything works, I think it's good to merge.

I ran the latest version successfully in my lab. Only the name of a variable was not correct (see last commit).

PR #391, opened 2021-10-05
Feature/add combell plugin

stevenvolckaert:
Added Combell Posh-ACME DNS plugin and documentation, for creating Let's Encrypt HTTPS certificates.

Tested with my Let's Encrypt account and it works, including wildcard certificates (`*.example.com`).

Your review and comments are appreciated!

### DIFF
--- a/Posh-ACME/Plugins/Combell-Readme.md
+++ b/Posh-ACME/Plugins/Combell-Readme.md
@@ -1,0 +1,36 @@
+# How To Use the Combell DNS Plugin
+
+This plugin works against the [Combell][1] DNS provider. It is assumed that you have already setup an account and created the DNS zone(s) you will be working against.
+
+[Combell NV][1] is a hosting provider based in Belgium. Besides offering hosting solutions, Combell NV is also an [ICANN Accredited Registrar under IANA number 1467](https://www.icann.org/en/accredited-registrars?sort-direction=asc&sort-param=name&page=1&iana-number=1467&country=Belgium).
+
+## Setup
+
+Using Simply.com API requires only your account name or account number and API Key which can be found on the [account](https://www.simply.com/en/controlpanel/account/) page.
+
+## Using the Plugin
+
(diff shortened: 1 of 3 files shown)


rmbolger:
Hey @stevenvolckaert. Thanks for the submission! A couple minor things before I get too deep into actually testing stuff.

- I recently started deprecating all of the Insecure parameter sets for plugins to prep for their removal in 5.x and I'm trying to avoid adding any new ones on new plugins.  So you can remove those and simplify things a bit.
- Typically for APIs that have both a Key and Secret, I only make the Secret a SecureString because the key is more like a username. It's not a huge deal if you'd rather keep them both SecureString values. It can just make it easier to distinguish between different credentials at a glance.

stevenvolckaert:
Thanks @rmbolger for your suggestions, I'll push an update later this week.

stevenvolckaert:
@rmbolger I've just pushed commit https://github.com/rmbolger/Posh-ACME/pull/391/commits/68b6e84bf74ab253fdf47badfc60be8767fa7983, can you have look?

Because we prefer not to save credentials in cleartext on our servers, I'd prefer to keep the `CombellApiKey` parameter of type `SecureString` .

rmbolger:
Got a pretty quick response from support saying that API access is only for resellers, no exceptions. I added a few minor tweaks. But otherwise, everything looks ok from a code standpoint. If you can add a warning to the guide about not being able to use the API unless you're a reseller, that would be great. Then I'm ready to merge as long as things are still working on your end.

rmbolger:
Sorry for the delay. Got busy with other stuff. That reseller trial page looks promising though. I'll check it out and review the guide changes when I get a chance.

stevenvolckaert:
Using the `POSHACME_PLUGINS` variable works great, thanks for this suggestion!

stevenvolckaert:
I've removed the `POSHACME_PLUGINS` variable and updated Posh-ACME to version 4.11.0; everything works.

Thanks again for all your support @rmbolger!

rmbolger:
Excellent!

PR #407, opened 2021-11-23
Fix example to use secure string param

webprofusion-chrisc:
- Example uses deprecated param name


rmbolger:
Ooh, thanks. I should double check the rest of the guides for deprecated params as well.

PR #429, opened 2022-03-19
Add Porkbun plugin and documentation

CaiB:
Adding a plugin to support https://porkbun.com DNS. 

I cannot test IDN support, as I don't have a domain that uses it, and their documentation isn't clear on what format the API expects.

Please let me know if there's any mistakes or anything I should change.

Thanks!

### DIFF
--- a/Posh-ACME/Plugins/Porkbun.ps1
+++ b/Posh-ACME/Plugins/Porkbun.ps1
@@ -201,7 +201,7 @@ function Get-PorkbunDomainInfo
         Write-Debug "Found domain `"$NameToCheck`"";
         return New-Object 'PSObject' -Property @{ Domain = $NameToCheck; Records = $ResultData.records; };
     }
-    throw "No matching domain could be found for `"$LongName`" on this Porkbun account. Check that the domain is correct, and that your API key and secret are entered correctly.";
+    throw "No matching domain could be found for `"$LongName`" on this Porkbun account. Check that the domain is correct, that your API key and secret are entered correctly, and that you've enabled API access for this domain in the settings.";
 
     <#
     .SYNOPSIS
--- a/Posh-ACME/Plugins/Porkbun.ps1
+++ b/Posh-ACME/Plugins/Porkbun.ps1
@@ -1,0 +1,223 @@
+using namespace Systen;
+using namespace Microsoft.PowerShell.Commands;
+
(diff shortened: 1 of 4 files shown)


rmbolger:
Hey @CaiB, thanks for the plugin submission. I picked up a cheap domain to test with, but I'm running into a problem. I've enabled API access on the domain, created myself a key, and started by running a `Publish-Challenge example.com` with Debug/Verbose enabled the appropriate set of arguments. It looks like the initial check against the `_acme-challenge.example.com` name is failing and the code never moves on to the other checks in the for loop.

```
DEBUG: Something went wrong while checking domain "_acme-challenge.zudifawe.xyz"
Invoke-WebRequest: C:\Users\rmbolger\Documents\PowerShell\Modules\Posh-ACME\Plugins\Porkbun.ps1:190
Line |
 190 |  … APIResult = Invoke-WebRequest -URI $RequestURL -Body $RequestBody -Me …
     |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | {"status":"ERROR","message":"Invalid domain."}
```

CaiB:
That's really bizarre, whenever I requested an "invalid" domain from the API, it gave me back a 4xx status code (hence checking for `InvalidOperationException`), I've never seen an actual "ERROR" API status...
I think the right way to go would be to change the `throw;` on line 195 to a `continue;`.

CaiB:
Hmm, I wonder if this is a result of different PowerShell versions handing 4xx status codes differently. Mine causes a `InvalidOperationException` and yours throws something else.

I’m using the default Windows PowerShell (not updated to PowerShell Core).

rmbolger:
I think you're right. Just tested against PS 5.1 and it works there.

rmbolger:
Pushed some changes to fix compatibility between PowerShell additions and tweaked how the actual queries were logged a bit. This is now working for me in both PS 5.1 and 7 in the usual tests I like to run. How does it work for you?

CaiB:
It appears to be working fine for me

CaiB:
I was able to set up my RDG server and test installation of a certificate on there using this, and it worked perfectly. Haven't tested renewal yet.

rmbolger:
Excellent. We'll call this done.

PR #436, opened 2022-04-21
Update Export-CertPfx.ps1

kgcurran:
update cert Subject property to SubjectDN


rmbolger:
Hi @kgcurran. Can you provide some context for this change? Why do you believe it is necessary? What problem does it solve?

PR #447, opened 2022-08-05
Use modern PBES2 for private keys in PFX (fixes #446)

### LINKED ISSUE #446: PFX files generated cannot be used by OpenSSL 3.x without enabling legacy mode
The version of BouncyCastle used generates a PKCS12 file that uses RC2-40-CBC for encrypting certificates, which isn't really useful for encryption these days - 40 bit RC2 isn't that strong.

OpenSSL 3.x hasn't made a new implementation, instead relegating it to the legacy implementation, which is opt-in.

This can be fixed by using `Pkcs12StoreBuilder` and telling it to use `PbeWithShaAnd3KeyTripleDesCbc` for certificates, instead of creating the `Pkcs12Store` directly.

Note that 3DES isn't exactly the best these days, but I don't know if BouncyCastle handles AES256 correctly in PFX, or how to use it.

ChlorideCull:
Fixes #446

This raises security a bit, allows OpenSSL 3.x to read it without legacy mode, and in general is just a good idea.

As for compatibility, it *should* be well supported. OpenSSL, as far as I know, supports it since 1.1.1, and all earlier versions are out of support upstream. Windows also has no issues with it.

### DIFF
--- a/Posh-ACME/Private/Export-CertPfx.ps1
+++ b/Posh-ACME/Private/Export-CertPfx.ps1
@@ -9,7 +9,8 @@ function Export-CertPfx {
         [string]$OutputFile,
         [string]$ChainFile,
         [string]$FriendlyName,
-        [string]$PfxPass=''
+        [string]$PfxPass='',
+        [switch]$UseModernPfxEncryption
     )
 
     $CertFile = $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath($CertFile)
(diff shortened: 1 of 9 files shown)


rmbolger:
My fear with making this the new default is that it will unintentionally break folks who might still be using OpenSSL 1.0.x for some reason. Might need to add support behind a flag associated with the order at least until the next major module version (at which point it can become the new default and the flag can revert to legacy mode).

ChlorideCull:
> My fear with making this the new default is that it will unintentionally break folks who might still be using OpenSSL 1.0.x for some reason.

Debian buster was the last version of Debian to ship 1.0.x, and the LTS support for that ended in June. I think the only major Linux distro still shipping 1.0.x is RHEL 7, which is two major versions behind. Given that 1.0.x is EOL from upstream as well, whoever is still relying on it might deserve a bit of a wake up call 😉

That being said, it's your call to decide :)

rmbolger:
I really do appreciate the work on figuring this out, @ChlorideCull. But out of an abundance of caution, I'm going to gate it behind a per-order opt-in flag until the next major version of the module. With the changes I just pushed, you can do any of the following:

```powershell
# Enable on new cert
New-PACertificate 'example.com' -UseModernPfxEncryption

# Enable on new order
New-PAOrder 'example.com' -UseModernPfxEncryption

# Enable on existing order. This will also re-write existing PFX files.
Set-PAOrder -UseModernPfxEncryption

# Disable on an existing order. This will also re-write existing PFX files.
Set-PAOrder -UseModernPfxEncryption:$false
```

Like most other order properties, `Submit-Renewal` will also carry over the flag to renewals as well.

Let me know how this works for you and I'll get it merged.

ChlorideCull:
Can confirm this works with your changes!

Issue #451, opened 2022-08-25
GCloud should make the DNS project/zone configurable.


lukeschlather:
In reading your docs I was a little bemused by the instructions to create a custom role - great in principle but in practice I'm not sure the "smaller" role really reduces the scope significantly from DNS Administrator.

However, it is possible to create a hosted zone in a separate project, and only grant your service account access to DNS administrator in that project, which is much more useful. You can even in principle create a hosted zone just for _acme-challenge.example.com and give it access to that.

Incidentally, I am using a service account in a separate project and spent several hours puzzling over why I was getting "403 Forbidden" until I realized that Posh ACME was incorrectly inferring the project from my service account's project and listing hosted zones in that project (rather than the DNS Project which contains my hosted zone.) I think the error messaging here could be better, but also I think explicitly configuring the project and hosted zone (maybe with an option to infer them) would be a better API.

i.e.

```
    $certificate = New-PACertificate -CSRPath $csr_file `
      -Plugin GCloud     `
      -PluginArgs @{GCKeyFile=$serviceAccountFile,GCProject=<myproject>,GCHostedZone=<zone>} `
      -DirectoryURL ZEROSSL_PROD `
      -verbose -debug
```

rmbolger:
Hey @lukeschlather, thanks for the suggestion! I actually had no idea you could do cross-project access like that. It's been a while since I looked at this particular plugin's code, so it's probably due for a refresh anyway. But it seems like it would be pretty straightforward to add an optional project argument. It could plausibly take an array so you could specify multiple projects.

> In reading your docs I was a little bemused by the instructions to create a custom role - great in principle but in practice I'm not sure the "smaller" role really reduces the scope significantly from DNS Administrator.

I mean...it's better than nothing given the options. While compromised credentials would still allow someone malicious to mess up your existing zones, at least they won't be able to create new zones (costing you money) or delete existing zones. Azure has everyone beat in their DNS permissions model except maybe OVH which allows permissions down to the individual record. Though that's pretty cumbersome to make work.

rmbolger:
Hey @lukeschlather, have a look at PR #452 and see if the [updated plugin file](https://github.com/rmbolger/Posh-ACME/blob/ba199ce84f98d6c1c604ee8bddf3c6a99209d6ef/Posh-ACME/Plugins/GCloud.ps1) works for you.

rmbolger:
This is now live in 4.15.0

PR #452, opened 2022-08-25
Update GCloud to support for querying additional projects

rmbolger:
Addresses issue #451 by adding a `GCProjectId` string array parameter that accepts one or more Google Cloud project IDs containing the zones needing to be modified by the plugin.

The parameter is optional and only required if

- The zones to modify don't reside in the same project as the service account reference by `GCKeyFile`
- The zones to modify reside in more than one project


lukeschlather:
I tested it and it works for my use case.

PR #456, opened 2022-09-25
Added Plugin for Active24 DNS provider

pastelka:
Hi, I have prepared a DNS-1 plugin for DNS provider from Czech Republic - Active24 ([www.active24.cz](http://www.active24.cz/)).


rmbolger:
Thanks @pastelka. I ordered a domain from them to test out the plugin and I'm just waiting for it to be activated.

rmbolger:
Everything works great, @pastelka. Thanks for the effort. I made a few tweaks to make sure things are unique between plugins. The `Token` param is now `A24Token` and there is no more insecure token param since it's not necessary on new plugins.

I'm ready to merge if everything still works for you.

pastelka:
@rmbolger Yes, It works.

PR #457, opened 2022-10-05
Fix for NameSilo API change

rkone:
NameSilo has made an undocumented API change to the listDomains query, it now returns the domain expiry as an xml property.  This filters out the expiry so the domain can be found again.


rmbolger:
Well that wasn't very nice of them. The fix seems simple enough. As the original plugin author, I'll trust you that it works. Merging.

PR #460, opened 2022-11-14
Implement Bunny.net DNS Provider

webprofusion-chrisc:
This pull request implements a DNS provider for https://bunny.net DNS zones.

### DIFF
--- a/Posh-ACME/Plugins/Bunny.ps1
+++ b/Posh-ACME/Plugins/Bunny.ps1
@@ -1,0 +1,237 @@
+function Get-CurrentPluginType { 'dns-01' }
+
+function Add-DnsTxt {
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory, Position = 0)]
+        [string]$RecordName,
+        [Parameter(Mandatory, Position = 1)]
+        [string]$TxtValue,
+        [Parameter(Mandatory, Position = 2)]
+        [securestring]$BunnyAccessKey,
+        [Parameter(ValueFromRemainingArguments)]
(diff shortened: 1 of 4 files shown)


rmbolger:
Nice! Thanks, @webprofusion-chrisc. I hadn't heard of these guys before. The no hassle trial made testing things super easy and everything seems to be working.

I pushed a minor change that adds `ErrorAction = 'Stop'` to the `$restParams` hashtables. Without it, Invoke-RestMethod won't actually throw terminating errors that would be caught by the try/catch blocks.

webprofusion-chrisc:
Thanks! Interestingly I did actually have trouble after this getting Let's Encrypt to actually validate using it. It was creating values for instances ABCD123XYZ then LE was saying that was the unexpected value, Trying with ZeroSSL now but they seem to have issues with orders getting stuck at Processing currently :) - obviously that's unrelated to validation, just a coincidence.

I'm also not sure if the record TTL is too high, or if that really matters or not - it's currently 5 mins and I think they support down to 15 seconds.

webprofusion-chrisc:
So some of the validation/order problems were caused by my test CAA record (which included the account specific restriction extension), so it stopped issuance but they just didn't all report that as the problem. Zero SSL is still hanging, BuyPass Go works, Let's Encrypt fails to validate, worth checking yourself.

rmbolger:
LE staging worked for me first try. At least for LE, the TTL shouldn't matter since they only cache results up to 60 sec unless the TTL is lower than that. Wouldn't hurt to whatever their minimum is though.

PR #467, opened 2022-12-13
Posh-ACME Docs: Added OVH Troubleshooting section

jan-oratowski:
I added a Troubleshooting section with one error I got when setting it up.

It turns out it was server misconfiguration, but the error message (from the OVH) wasn't very helpful, which made me waste some time to figure it out.


rmbolger:
Thanks for the PR, @jan-oratowski. I might tweak the wording a bit before merging it, but just wanted to let you know I saw it if I don't get to it right away.

PR #476, opened 2023-02-07
Add Ilait DNS plugin

etfz:


rmbolger:
Hi @etfz. Thanks for the submission. I'm a little behind on reviewing new plugins right now. I have a couple questions about this provider though. It looks like they're sort of a provider for other businesses to resell services for and they don't cater to individual consumers. So it seems like I won't be able to get an account to test with directly. Is that right? Are there other providers you know of that use Ilait's API that I could potentially get an individual account with?

etfz:
Hi,

No worries! I have already deployed it locally.

Yes, I do believe their platform is closed in that regard, indeed. I suppose that I could create a sandbox customer account for you, if you like. Where could I send the details?

rmbolger:
Nice. A sandbox account should work just fine. You can send details to `ryan-oss` (at) `xyto.cc`.

etfz:
Apparently customer accounts do not have API access, unfortunately, so uhh.. don't know what you want to do with this.

rmbolger:
Well that seems like a silly decision on their part. We can just close it and re-open if they ever open up access.

PR #477, opened 2023-02-09
Implement Google Domains plugin

webprofusion-chrisc:
This plugin implements a provider specifically for DNS based acme challenges via Google Domains.

### DIFF
--- a/Posh-ACME/Plugins/GoogleDomains.ps1
+++ b/Posh-ACME/Plugins/GoogleDomains.ps1
@@ -7,45 +7,47 @@ function Add-DnsTxt {
         [string]$RecordName,
         [Parameter(Mandatory, Position = 1)]
         [string]$TxtValue,
-        [Parameter(ParameterSetName="Single", Mandatory, Position=2)]
-        [string]$RootDomain,
-        [Parameter(ParameterSetName="Single", Mandatory, Position=3)]
-        [securestring]$AccessToken,
-        [Parameter(ParameterSetName="MultipleInsecure", Position=2)]
-        [hashtable]$DomainTokensInsecure,
+        [Parameter(Mandatory)]
+        [pscredential[]]$GDomCredential,
         [Parameter(ValueFromRemainingArguments)]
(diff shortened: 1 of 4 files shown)


rmbolger:
Hey @webprofusion-chrisc, is this API public enough to test against for non-insiders yet? Or am I gonna need to wait a bit more?

webprofusion-chrisc:
Not yet, ping me an email to christopher.cook {at} webprofusion.com and I'll get back to you there!

rmbolger:
Reworked things to use an array of PSCredential objects as the primary parameter which is now supported in the main branch. Each PSCredential provided must have the username be a root domain and the password as that domain's access token. You'll need one cred for every unique domain represented in a cert order.

If you want to test more, you can either rebase your branch against the upstream main or I could just merge this PR and you can test against the main branch directly.

rmbolger (approved):

webprofusion-chrisc:
Cool, I see the changes you mean for the PSCredential array, I'll need to tweak how Certify The Web calls this but that should be fine. Feel free to merge it and I'll let you know if I hit any problems with it.

webprofusion-chrisc:
Works fine, CTW change was tiny as we already use PSCredential for a few other plugins.

rmbolger:
This is SO close to the perfect ACME specific API because a single call does what is necessary for each action (add/remove) without regard for the current state of the zone. It looks like you could even theoretically skip the remove action if you set the `keepExpiredRecords` flag to false. There are only two hassles which I'm sure are too late to get the Google team to change course on.

 - Needing to know the root domain name (or any unique zone ID). Knowing just the FQDN I want to add should be sufficient.
 - Having the authentication token be zone specific instead of account specific or specific to a configurable set of zones. I should be able to provide a single credential and update all of the zones I care about.

webprofusion-chrisc:
Yeah I agree with these, I think they affect Posh-ACME a little more than they do Certify but I'll pass on your feedback to them.

PR #479, opened 2023-02-20
Fixed bug in New-Csr splat

ZamElek:
Fixed a minor bug in the function Submit-OrderFinalize


rmbolger:
Hi @ZamElek. Thanks for reaching out, but the code is correct as-is. The Order object isn't meant to be splatted to `New-Csr`. It takes a whole PAOrder object as a parameter.

https://github.com/rmbolger/Posh-ACME/blob/9e690b9a180d70b794bb0a2ace6d671fca0bd0fb/Posh-ACME/Private/New-Csr.ps1#L1-L8

Did you run into a problem that this change seemed to fix?

ZamElek:
Sorry, I had an issue with my certificate renewal process. I started to debug the module and found that the issue was with New-Csr parameters. I saw that hashtable pases as variable to the command and it was strange for me :). And after I changed it to the splat - the error has gone... Sorry, for that, next time I will double chech before submitting the pull request.

PR #485, opened 2023-03-29
Fix broken DDNSNameserver param in RFC2136 plugin

gvengel:
The recent change (a2ff71c) to make the `DDNSNameserver` param optional in the RFC2136 plugin, actually disabled the ability to specify the DNS server manually, since the check to see if the param had been provided was using the wrong variable name. This forced `nsupdate` to always try to detect it via SOA, which could be wrong if people are using split DNS, etc.


rmbolger:
Whoops. Guess I forgot to test the original functionality in addition to the changed functionality. Thanks for the fix, @gvengel. I also copied it to the Remove function as well. I'll get this merged and a new release out asap.

Issue #418, opened 2022-01-27
AWS Lambda runtime support


frankvdbh:
We are trying to use Posh-ACME in an AWS Lambda function.
The lambda environment uses .Net core 3.1.
We package the lambda function using AWS Powershell tools into a lambda package. During this stage it is bundling the Posh-ACME module (
https://docs.aws.amazon.com/lambda/latest/dg/powershell-handler.html )
However when the lambda is being executed, it complains on importing one of the module functions and I'm not able to find out the reason. We assume something specific to the Amazon Linux base image but wonder why it complains on this specific function. 
Anyone else tried te same ? 
We tried a more manual bundling approach where we locally extract the module files, leave out the Revoke-PACertificate function, and this does work.

The error output is below:
```
START RequestId: 504063a8-436b-4b65-a860-3629e7e7787d 
Version: $LATEST
Importing module ./Modules/Posh-ACME/4.12.0/Posh-ACME.psd1

[Error] - Failed to import function /var/task/Modules/Posh-ACME/4.12.0/Public/Revoke-PACertificate.ps1: Could not find a part of the path '/var/runtime/ref'.

[Error] - Cannot bind argument to parameter 'Path' because it is null.

[Error] - Value cannot be null. (Parameter 'The provided Path argument was null or an empty collection.')
[Error] - Cannot bind argument to parameter 'Path' because it is null.
[Error] - Cannot bind argument to parameter 'Path' because it is null.
```

rmbolger:
Hey @frankvdbh. Thanks for reaching out. Unfortunately, I don't have much experience with Lambda and I don't recall anyone posting about similar issues in the past.

That said, I do have an AWS account that I could poke around on and try some things to see if I can at least reproduce the problem.

rmbolger:
This is now fixed in 4.18.0

PR #489, opened 2023-04-19
Fix for issue #418. Change Add-Type to native enum

garthmccormack:
Hi @rmbolger 

I ran into the problem described in Issue #418. After investigating I identified the part causing the error to be the Add-Type call to add the enum for RevocationReasons. I tested replacing this with the native enum keyword which worked. Enum was introduced in Powershell v5 so this should meet your Powershell version requirements.

As it happens the documentation uses RevocationReasons which the new enum matches as well so the only changes are in the Revoke-PACertificate.ps1 file.

### DIFF
--- a/Posh-ACME/Public/Revoke-PACertificate.ps1
+++ b/Posh-ACME/Public/Revoke-PACertificate.ps1
@@ -14,7 +14,7 @@ function Revoke-PACertificate {
         [string]$CertFile,
         [Parameter(ParameterSetName='CertFile',ValueFromPipelineByPropertyName)]
         [string]$KeyFile,
-        [PoshACME.RevocationReasons]$Reason,
+        [RevocationReasons]$Reason,
         [switch]$Force
     )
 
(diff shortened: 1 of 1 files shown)


rmbolger:
Ooh, nice find @garthmccormack! The `Add-Type` was indeed probably a leftover from before I started requiring PSv5 minimum. I still don't quite get why the Add-Type call causes Lambda to break. But if this fixes it and doesn't break anything else, I'll take it.

Give me some time to poke at it and I'll get it in.

garthmccormack:
I had a quick look at what the Add-Type command actually does in version 7 of Microsoft.PowerShell.Utility and it looks like the issue is in this bit from the .Net dll file:
```
AddTypeCommand.s_netcoreAppRefFolder = Path.Combine(Path.GetDirectoryName((Assembly.GetEntryAssembly() ?? typeof(PSObject).Assembly).Location), "ref");
```
The expectation from the above code is that it would resolve the EntryAssembly to somewhere like C:\Program Files\PowerShell\7 on a windows system which then has a "ref" folder under it. However, when running on AWS Lambda the script is wrapped with their own launcher using dotnet6 as the runtime. The entry assembly then sits in /var/runtime which doesn't have the ref subfolder. 

Just tested and the EntryAssembly for Lambda is /var/runtime/Amazon.Lambda.RuntimeSupport.dll and C:\Program Files\PowerShell\7\pwsh.dll on Powershell 7.

rmbolger:
I now recall why I had originally gone with the `Add-Type` route instead of moving to the native PowerShell enums. With `Add-Type` the enum becomes available for use directly within the callers scope. But with the native enum, that no longer works which is a known issue with both native enums and classes. The available workarounds to get it into the caller's scope are cumbersome and have caveats.

However, despite the lack of direct access, tab completion of the enum values for `Revoke-PACertificate -Reason <tab>` works using both methods. And even without it, interpretation of the enum value names without being fully qualified works just fine in both PS 5.1 and 7+.

```powershell
# this is way easier to type and read
Revoke-PACertificate -Reason keyCompromise

# than this
Revoke-PACertificate -Reason ([PoshACME.RevocationReasons]::keyCompromise)
```

Since it's unlikely any other software outside of this module is depending on using the enum, losing access to it from outside the module scope is not likely to negatively affect anyone. So I'm going to sneak it in to a 4.x release rather than bumping the major version number.

Thanks again for research and fix.

PR #492, opened 2023-05-11
Update Simply.md

tomsommer:
The name of the brand is "Simply.com"


rmbolger:
It would be easy enough to add a clone of the plugin called SimplyCom. But users would have to manually migrate their cert configs to it. The old one would also need to stay in place until the next major version to avoid breaking folks.

rmbolger:
Apologies. This is now live on the doc site. I'll work on the plugin clone when I have more time.

PR #493, opened 2023-05-11
Add support for Hurricane Electric's DynDNS to update TXT-record

jbrunink:
Just a WIP. With this you can use the DynDNS mechanism to update  TXT records for dns.he.net.

What are your thoughts?


rmbolger:
@Hey @jbrunink, I love it. The web-scraping method the current plugin uses always left a bad taste in my mouth. If the DynDNS method can work using the same parameters as the original plugin without any limitations, I'd even say it we should just replace the current implementation rather than creating an alternate plugin.

rmbolger:
What sort of manual intervention?

PR #494, opened 2023-05-14
Add support for Hurricane Electric's Dynamic TXT API to update TXT-record 

jbrunink:
I've added support to use Hurricane Electrics Dynamic TXT API. For some reason the other PR got closed by me, *shrug*

Thanks for the good work you're doing on the project!


rmbolger:
Sorry this took a while, @jbrunink. I took the liberty of changing up a few things like shortening the plugin name to `HurricaneElectricDyn` and changing the name of the credential param to `HEDynCredential` so it doesn't conflict with the other plugin. I also added support for multiple credentials so you can get SAN certs with multiple names or wildcards. The user guide goes into detail about how you need to create the TXT record first and generate a key/password for each one.

In any case, it seems to be working for me so I'm gonna merge it and get it into the next release. Let me know if you have any problems with it.

PR #546, opened 2024-04-29
Fix #545 - Replace ternary operator for PS 5.1 compatibility

### LINKED ISSUE #545: OVH plugin is not compatible with PowerShell 5.1
The OVH plugin uses the ternary operator which is invalid in PowerShell 5.1. You can reproduce the error by running `Get-PAPlugin -Plugin OVH -Params` in Windows PowerShell. I'm submitting a PR now and will reference this issue.

I don't actually use OVH - I only discovered it because I was writing some code to automatically prompt users for parameters based on the plugin they're using, and I wanted to see all the plugin argument types. While investigating all the possible parameter types for all plugins, I ran this and got a splash of red from `OVH.ps1`.

```powershell
$plugins = Get-PAPlugin
$plugins.Name | % { Get-PAPlugin $_ -Params } | Group-Object ParameterType
```

joshooaj:
This PR fixes #545 by replacing the ternary operator in a `Write-Debug` string-formatting statement. 

While this change makes the file compatible with PowerShell 5.1, I don't have a way to test the OVH plugin and verify the output of the `Write-Debug` statement is shown correctly with/without a value for `$Body`.

### DIFF
--- a/Posh-ACME/Plugins/OVH.ps1
+++ b/Posh-ACME/Plugins/OVH.ps1
@@ -408,7 +408,7 @@ function Invoke-OVHRest {
     # add the body if there is one
     if ($Body) { $restArgs.Body = $Body }
 
-    Write-Debug ('{0} {1}{2}' -f $restArgs.Method,$restArgs.Uri,(($Body) ? "`n$Body" : ''))
+    Write-Debug ('{0} {1}{2}' -f $restArgs.Method,$restArgs.Uri,"$($restArgs.Body)")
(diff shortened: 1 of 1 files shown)


rmbolger:
Whoops. Thanks. The ternary was only there to make the line break conditional on the existence of the body. As a debug line, it doesn't really matter though. I'd probably just hard code the line break in the source string after changing from single to double quotes.

```powershell
"{0} {1}`n{2}" -f $restArgs.Method,$restArgs.Uri,$restArgs.Body
```

PR #547, opened 2024-05-03
Add account limits note in GoDaddy.md

webprofusion-chrisc:
According to GoDaddy API support there are new limits for API access:

> We have recently updated the account requirements to access parts of our production Domains API. As part of this update, access to these APIs are now limited:
> 
>     Availability API: Limited to accounts with 50 or more domains
>     Management and DNS APIs: Limited to accounts with 10 or more domains and/or an active Discount Domain Club plan.
> 
> If you have lost access to these APIs, but feel you meet these requirements, please reply back with your account number and we will review your account and whitelist you if we have denied you access in error.
> 
> Please note that this does not affect your access to any of our OTE APIs.
> 
> If you have any further questions or need assistance with other API questions, please reach out.

### DIFF
--- a/docs/Plugins/GoDaddy.md
+++ b/docs/Plugins/GoDaddy.md
@@ -2,7 +2,9 @@ title: GoDaddy
 
 # How To Use the GoDaddy DNS Plugin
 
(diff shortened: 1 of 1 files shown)


rmbolger:
Well that is a lame decision on their part. I wonder if they also broke the ability to use their DNS services without actually having the domain registered with them.

In any case, thanks for the update.

PR #550, opened 2024-05-27
update DomainOffensive plugin

henrikalves:
Hi,
we are moving our API to a different subdomain (the old one will continue to work for now),
I also have updated the links to the documentation


rmbolger:
Awesome. Thanks. I'll get this out into a new release soon.

PR #553, opened 2024-06-21
Add TencentDNS plugin (DNSPod using Tencent Cloud API)

xiaotiannet:
add plugins DnsPod_v3 . used dnspod api 3.0 .
because old version Coming off the line.
see https://docs.dnspod.cn/api/


rmbolger:
Hi @xiaotiannet. Thank you for the PR and the effort on building API 3.0 compatibility. Do you know when the old version go offline?

It looks like API 3.0 can be used both with Tencent API keys and old DNSPod API keys. I was thinking of trying to incorporate your API 3.0 changes into the existing plugin so that folks still using the old DNSPod keys would not need to change anything.

rmbolger:
After doing a bit of testing, the API 3.0 doesn't seem to like the old DNSPod Tokens even if I create a new one from scratch. So perhaps keeping this as a new plugin makes more sense. I think I'd like to preemptively rename it to `TencentDNS` though.

xiaotiannet:
> Hi @xiaotiannet. Thank you for the PR and the effort on building API 3.0 compatibility. Do you know when the old version will go offline?
> 
> It looks like API 3.0 can be used both with Tencent API keys and old DNSPod API keys. I was thinking of trying to incorporate your API 3.0 changes into the existing plugin so that folks still using the old DNSPod keys would not need to change anything.

thanks for you replay。 Im sorry,I don't know  when old version will offiline . just see this  https://docs.dnspod.cn/api/api3/

> 
> Will the DNSPod API (old version API) be maintained in the future?
> The DNSPod API will no longer be maintained in the future, and the old version of the DNSPod API will no longer have service commitments. Please switch to Tencent Cloud DNSPod API 3.0 as soon as possible based on actual usage. API 3.0 entry


and  yes . API 3.0 doesn't  woking with  DNSPod Tokens , and DNSPod  doesn't  woking with 3.0 Key , They are independent specifications 。

PR #579, opened 2024-11-04
Fix: Modify domain data extraction method to handle PSCustomObject format with numeric keys

radmax:
## Description
This PR updates `$zones` assignment to handle WEDOS API responses where `domain` is structured as a `PSCustomObject` with numeric keys.

## Changes
- Replaced `Select-Object -Expand domain` with `.PSObject.Properties | ForEach-Object { $_.Value }` to handle domains with numeric keys.
- Added debug output to confirm `$zones` contains correct domain data.

### DIFF
--- a/Posh-ACME/Plugins/WEDOS.ps1
+++ b/Posh-ACME/Plugins/WEDOS.ps1
@@ -260,8 +260,15 @@ function Find-Zone {
     }
 
     # Get all of the domains on the account
-    $zones = Invoke-Wedos dns-domains-list $Credential | Select-Object -Expand domain
-    if (-not $zones) {
+    $wedosResponse = Invoke-Wedos dns-domains-list $Credential
+    
+    if ($null -ne $wedosResponse.domain) {
+        # Extract each domain by accessing the properties of the PSCustomObject
+        $zones = @($wedosResponse.domain.PSObject.Properties | ForEach-Object { $_.Value })
+    
+        # Debug output to confirm zones
+        Write-Debug "Found domains: $($zones | ConvertTo-Json -Depth 10)"
+    } else {
         Write-Warning "No WEDOS hosted domains found."
         return
     }


radmax:
Hello, I was wondering the same. But my response is:

```
DEBUG: Response:
{
  "response": {
    "code": 1000,
    "result": "OK",
    "timestamp": 1730886863,
    "clTRID": "Posh-ACME 2dbbf1c5-c6cb-47de-8eb1-30bad17f97b7",
    "svTRID": "1730886863.9062.21431.1",
    "command": "dns-domains-list",
    "data": {
      "domain": {
        "24": {
          "name": "**domain**.com",
          "type": "primary",
          "status": "active"
        },
        "11": {
          "name": "**domain**.cz",
          "type": "primary",
          "status": "active"
        },
        "16": {
          "name": "**domain**.eu",
          "type": "primary",
          "status": "active"
        },

...etc.
```

rmbolger:
@radmax I made some tweaks that should hopefully allow for either response type to work. Can you give the updated version a try on your account and see if it still works for you?

radmax:
Will do. I also tried to contact their helpdesk, if they can explain this behavior. They asked if i can give them request, that returned array, just as one you put in this conversation. 

Can i take it and forward it to them?

rmbolger:
The copy is posted is modified a bit to hide the real domains, but sure.

radmax:
Hello, i tested updated code and it works fine. I also again checked Wedos KB and they have there example with data returned as object, so i guess it should be standard response. https://kb.wedos.com/en/wapi-api-interface/wdns-en/wapi-wdns/#dns-domains-list 

Anyway, thanks for great library and your help with resolving my problem 😉.

rmbolger:
Hah, that's the same docs link I posted last week. And when I did, I swear it still said the response would be an array of objects. I wonder if they stealth updated it. In any case, my own account's responses are still returning an array. Not sure it's worth dealing with support about it though. The code that works for both should be sufficient unless they muck with the responses again.

Issue #486, opened 2023-04-04
EfficientIP SOLIDServer integration


amckinney92:
Solidserver DDI appliances have REST API functionality and could integrate with PoshACME for a certificate solution that leverages in house DNS to modify records on the fly for validation.

https://gitlab.com/efficientip/solidserverrest/-/tree/master/

tall27:
I also was looking for the same integration. 
Please consider to add it to Posh-ACME module.
thank you!

rmbolger:
Hey @amckinney92 and @tall27, thanks for reaching out. The project you linked appears to be a python library for interacting with SOLIDServer's existing REST API. It might be useful as guidance for developing a Posh-ACME plugin, but I can't use it as-is. 

What would be useful is an instance of SOLIDServer I can test against and a link to their REST API documentation. Without an existing test instance, it's going to take a while because I'll have to spin up my own instance which will involve contacting their sales team to request a free trial. Then, I'll have to figure out where to deploy it and learn how to actually deploy it.

jamiekowalczik:
I just created a simple Powershell module for basic CRUD functionality to the DNS module of EfficientIP SOLIDServer.  I might be able to assist with this.

rmbolger:
SOLIDServer plugin is now live in 4.28.0

PR #582, opened 2024-11-08
Add EfficientIP SOLIDServer plugin

jamiekowalczik:
This is my first pull request, I apologize if I am doing it incorrectly.  As recently previously stated in your Issues by someone else, the quality and documentation you have put into this project is truly amazing.


rmbolger:
Just a heads up. I reached out to EfficientIP to see if I could procure some sort of trial/demo. Still waiting on a response.

rmbolger:
Adding a quick link to #486 which this PR would address.

jamiekowalczik:
Thank you for contacting them.  I hope they see the value in it.

rmbolger:
I've still heard nothing but crickets from EfficientIP about a demo/trial. I'm tempted to reach out from my work address, but that probably wouldn't happen until the new year.

Might you have access to installation media that could somehow find its way to me? I don't know how licensing works, so it may be pointless if it doesn't work without some sort of keys though. We run Infoblox at work which uses 60-day trial licenses by default. So I'm hoping there's something like that baked in.

alexissavin:
@rmbolger could be get in touch privately ?

I can certainly share access to a temporary public SOLIDserver instance.

Kind regards

jamiekowalczik:
Thank you so much Alexis! The SOAP/API guide will be a good doc to share with Ryan as well. I really appreciate you helping out with this request!

rmbolger:
@alexissavin Sure. my email is ryan-oss {at} xyto.cc.

rmbolger:
Hey @jamiekowalczik. Just letting you know @alexissavin got me access to a test instance and I've started digging in a bit. One thing we discussed on our call was likely renaming the plugin to `SOLIDServer` instead of `EfficientIP` to align better with their branding of the product. "EfficientIP" will still be in the docs though.

Other things I'm thinking about as I start poking around:

- Probably going to combine the username/password params into a single PSCredential param.

- Going to try and implement their Token authentication option if I can. The tricky part is that it requires SHA3_256 hashing support which only has native PowerShell support in in very recent PowerShell versions and [relatively recent versions of Windows 11 and Linux](https://learn.microsoft.com/en-us/dotnet/standard/security/cross-platform-cryptography#sha-3). We'll see how that goes.

jamiekowalczik:
That all sounds great to me. I trust your direction and will adapt. Thank
you for supporting SOLIDServer EfficientIP!

Posh-ACME/Plugins/EfficientIP.ps1
```diff
@@ -0,0 +1,237 @@
...
+    $EIPPasswordInsecure = [pscredential]::new('a',$EIPPassword).GetNetworkCredential().Password
+
+    $Endpoint = "dns_rr_add"
+    $queryParams += @{rr_name = $RecordName}
+    $queryParams += @{value1 = $TxtValue}
+    $queryParams += @{rr_type = "TXT"}
+    $queryParams += @{rr_ttl = "1"}
+
+    $queryString = ($queryParams.GetEnumerator() | ForEach-Object { "$($_.Key)=$($_.Value)" }) -join '&'
+    $Parameters = "dnsview_name=$($EIPView)&add_flag=new_only&check_value=yes&dns_name=$($EIPDNSName)&$($queryString)"
```
alexissavin, line 34: The view is optional - If not specified the dnsview_name should not be included in the request

Posh-ACME/Plugins/EfficientIP.ps1
```diff
@@ -0,0 +1,237 @@
...
+    )
+
+    # convert the secure secret to a normal string
+    $EIPPasswordInsecure = [pscredential]::new('a',$EIPPassword).GetNetworkCredential().Password
+
+    $Endpoint = "dns_rr_add"
+    $queryParams += @{rr_name = $RecordName}
+    $queryParams += @{value1 = $TxtValue}
+    $queryParams += @{rr_type = "TXT"}
+    $queryParams += @{rr_ttl = "1"}
```
alexissavin, line 31: Not sure why specifying a TTL of 1 ?

Posh-ACME/Plugins/EfficientIP.ps1
```diff
@@ -0,0 +1,237 @@
...
+   process {
+      $basicAuthValue = [Convert]::ToBase64String([Text.Encoding]::ASCII.GetBytes(("{0}:{1}" -f $EIPUsername, $EIPPassword)))
+      $Headers = @{
+         "Authorization"="Basic $basicAuthValue"
+         "Content-Type"="application/json"
+         "Accept"="application/json"
+         "charset"="utf-8"
+      }
+      $URI = "/rest/$($Endpoint)?$($Parameters)"
+      $URL = "https://$($EIPHostname)$($URI)"
```
alexissavin, line 187: Notice that unfortunately many still rely on self signed certificate. I hope it's possible to disable SSL certificat and hostname checks.

Posh-ACME/Plugins/EfficientIP.ps1
```diff
@@ -0,0 +1,237 @@
...
+      [Parameter(Mandatory,Position=1)][String]$EIPUsername,
+      [Parameter(Mandatory,Position=2)][String]$EIPPassword,
+      [Parameter(Mandatory,Position=0)][String]$Name,
+      [Parameter(Mandatory,Position=0)][String]$TxtValue,
+      [Parameter(Mandatory,Position=1)][String]$DNSName,
+      [Parameter(Mandatory,Position=2)][String]$View
+   )
+   Try{
+      $Endpoint = "dns_rr_list"
+      $Parameters = "WHERE=(rr_full_name='$($Name)'+AND+value1='$($TxtValue)')+AND+(dns_name='$($DNSName)'+AND+dnsview_name='$($View)')"
```
alexissavin, line 221: As the plugin only leverage TXT, maybe consider specifying the rr_type='TXT'

alexissavin (commented):
Valuable proposal :)

I just made some comments in the code that I hope will be useful.

Additionally, one thing could be improved: validate that the DNS server specified can actually be used to push the change. Ideally write operation should only be done on the smart, or one standalone server. This can be achieved by considering the value of the vdns_parent_id property of all objects server/view/zone/rr.

Kind regards

rmbolger:
Thanks for the suggestions, @alexissavin. In my local copy, I've already added a way to make cert validation and the View optional. I was considering trying to make the Server/dns_name parameter optional as well and find the zone via one or more `dns_zone_list` calls.

Regarding the parent id property, would it be safe to only allow zones where `vdns_parent_id = 0` indicating the parent is either not managed by Smart or is Smart itself?

rmbolger:
Interesting observation as I was messing with stuff tonight. If you do a `dns_rr_add` with no dns server or view specified and happen to have two copies of a zone that would match (like an internal and external copy), the add will succeed and create the record in both copies of the zone. But the response from the API will only return the `ret_oid` value of the last one created.

Seems like a potential bug. Feels like it should either throwing an error when the add parameters would be make the destination zone ambiguous or return all IDs of the records created.

rmbolger:
I managed to get token auth working despite the SHA3-256 library limitations by falling back to BouncyCastle for environments that don't support the .NET native version. You can basically just use your API Token+Secret as the username and password for the `SolidCredential` parameter as long as you also set the `SolidTokenAuth` switch as well. The available parameters now look like this:

```
Parameter       Type            IsMandatory
---------       ----            -----------
SolidCredential PSCredential    True
SolidAPIHost    String          True
SolidDNSServer  String          False
SolidView       String          False
SolidIgnoreCert SwitchParameter False
SolidTokenAuth  SwitchParameter False
```

I'm pretty happy with the overall state of the plugin at this point. Now I just need to write the user guide.

rmbolger:
Before I merge this, @jamiekowalczik can you test the updated version in your environment and let me know if everything still works for you? Also pinging @amckinney92 and @tall27 who were also interested in this over in #486. I'd love to get more testing in as many real environments as I can.

jamiekowalczik:
Thanks @rmbolger.  I confirmed that username/password as well as token auth both work well in my environment!

alexissavin:
> Thanks for the suggestions, @alexissavin. In my local copy, I've already added a way to make cert validation and the View optional. I was considering trying to make the Server/dns_name parameter optional as well and find the zone via one or more `dns_zone_list` calls.
> 
> Regarding the parent id property, would it be safe to only allow zones where `vdns_parent_id = 0` indicating the parent is either not managed by Smart or is Smart itself?

Yes, it seems safe to say that only zones with vdns_parent_id = 0 should be editable.

> The goal is that the View and Server only become mandatory if the user's environment has multiple copies of the same zone (like Internal and External) that would make picking one automatically ambiguous.

Sound good to me. However, any ambiguity should stop the process to prevent any impact on a production.

alexissavin:
> Interesting observation as I was messing with stuff tonight. If you do a `dns_rr_add` with no dns server or view specified and happen to have two copies of a zone that would match (like an internal and external copy), the add will succeed and create the record in both copies of the zone. But the response from the API will only return the `ret_oid` value of the last one created.
> 
> Seems like a potential bug. Feels like it should either throwing an error when the add parameters would be make the destination zone ambiguous or return all IDs of the records created.

Well the bug is not the one you point out :) The call is designed to create the RR in all matching zones if no server is specified. However, it should return all created objects OID. I will look in the backlog if I find any occurrence of this one.

alexissavin:
Many thanks to the both of you @jamiekowalczik @rmbolger.

PR #596, opened 2025-01-29
Update Azure.ps1 | Support for Linux CertThumbprints

Eric2XU:
Believe it or not, there is a way to add certs to Linux systems using PowerShell. However, this method does not present the PS-drive provider Microsoft.PowerShell.Security\Certificate. So the current Azure plugin code checking if the cert is there using cert:\ doesn’t work. I left existing code and just added a "well you didn’t find it, try these two new methods" sort of code. Technically I am pretty sure my method would work in Windows so it might be worth in the future swaping to just that but I cant fully test it. Anyone is welcome to take what I did and make it better. Anyone wanting to import a cert in Linux can use this example code: 

```
#import the PFX to your machines cert store
$StoreName = [System.Security.Cryptography.X509Certificates.StoreName]::My  
$StoreLocation = [System.Security.Cryptography.X509Certificates.StoreLocation]::CurrentUser  
$Store = [System.Security.Cryptography.X509Certificates.X509Store]::new($StoreName, $StoreLocation)  
$Flag = [System.Security.Cryptography.X509Certificates.X509KeyStorageFlags]::Exportable 

# Prompt for the password
$password = Read-Host -Prompt "Enter the password for the PFX file" -AsSecureString

$Certificate = [System.Security.Cryptography.X509Certificates.X509Certificate2]::new("./something.pfx", $password, $Flag)  $Store.Open([System.Security.Cryptography.X509Certificates.OpenFlags]::ReadWrite)  $Store.Add($Certificate) 
$Store.Close()

write-host "Certificate installed: " $Certificate.thumbprint
```

### DIFF
--- a/Posh-ACME/Plugins/Azure.ps1
+++ b/Posh-ACME/Plugins/Azure.ps1
@@ -547,13 +547,52 @@ function Connect-AZTenant {
         if ('CertThumbprint' -eq $PSCmdlet.ParameterSetName) {
             Write-Debug "Looking for cert thumbprint $AZCertThumbprint"
             # Look up the cert based on the thumbprint
+            $cert = $null
+            
             # check CurrentUser first
-            if (-not ($cert = Get-Item "Cert:\CurrentUser\My\$AZCertThumbprint" -EA Ignore)) {
+            $cert = Get-Item "Cert:\CurrentUser\My\$AZCertThumbprint" -EA Ignore
+            
+            if (-not $cert) {
                 # check LocalMachine
-                if (-not ($cert = Get-Item "Cert:\LocalMachine\My\$AZCertThumbprint" -EA Ignore)) {
-                    throw "Certificate with thumbprint $AZCertThumbprint not found in CurrentUser or LocalMachine stores."
-                }
+                $cert = Get-Item "Cert:\LocalMachine\My\$AZCertThumbprint" -EA Ignore
             }
+            
+            ### 
+            # This is the new method I added in to support linux systems
+            # its entirely possible this method would work for Windows and if so may be better as its not dependant on the cert: PS-Drive provider
+            # I dont know enough to say for sure though so I just add this below, only triggers if $cert wasnt found already by previous methods
+
+            $CurrentErrorAction = $ErrorActionPreference
+            
+            if (-not $cert) {
+                # Setting Error Action to Ignore to prevent the error message from being displayed will set back after block
+                $ErrorActionPreference = "Ignore"
+            } 
+
+            if (-not $cert) {
+                # check CurrentUser .NET Style
+                $store = New-Object System.Security.Cryptography.X509Certificates.X509Store("My", "CurrentUser") 
+                $store.Open([System.Security.Cryptography.X509Certificates.OpenFlags]::ReadOnly)
+                $cert = $store.Certificates.Find([System.Security.Cryptography.X509Certificates.X509FindType]::FindByThumbprint, $AZCertThumbprint, $false)
+                $store.Close()
+            }
+
+            if (-not $cert) {
+                # check LocalMachine .NET Style
+                $store = New-Object System.Security.Cryptography.X509Certificates.X509Store("My", "LocalMachine")
+                $store.Open([System.Security.Cryptography.X509Certificates.OpenFlags]::ReadOnly)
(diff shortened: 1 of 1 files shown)


rmbolger:
Interesting. I swear the last time I looked at this, the X509Store related stuff didn't work on Linux....or maybe it just didn't fully work? It would have been way back in PowerShell 6/.NET Core 2 days though. This [SO question](https://stackoverflow.com/questions/49789841/net-core-x509store-on-linux) seems to imply maybe the LocalMachine store is read-only.

Out of curiosity, what do you have that actually put your Azure auth cert into the .NET cert store on Linux? Or is that just what the native Azure modules do now?

In any case, I'll try to find some time to muck around with this. More options for non-Windows is always good.

Eric2XU:
I just tested it with su rights and I your right, I get this error when trying to add to the localmachine store:

> MethodInvocationException: Exception calling "Open" with "1" argument(s): "Unix LocalMachine X509Stores are read-only for all users."

I was being presumptuous that since CurrentUser worked that so would LocalMachine. 

This is my first PR so I have no idea what the best way would be to edit the code and resubmit (I would just remove the block for localMachine). 

I discovered this method (of creating a currentUser store) as part of the process of just geting connect-azAccount working. I had to manually make the PFX, upload to Azure portal under the SPN app reg, then import using the code mentioned in the PR. Once set, the Connect-azAccount works great. That is when I discovered it didnt work in the plugin, hence the PR to make it so. I did test the changes locally and it did work for me using currentUser. 

Hope that helps!

rmbolger:
For the purposes of this change, the LocalMachine store being read-only might still be ok since the code is only reading the certs. In general though, updating a PR's code is just a matter of adding commits to the repo/branch the PR is referencing.

In any case, I'm still curious about how the LocalMachine store functions. One of those SO answers claims it's looking in the CentOS default cert locations that normally hold PEM-formatted files:
- /etc/pki/tls/certs
- /etc/pki/tls/private

So I wonder if just manually copying your cert files to the right place would effectively add them to the LocalMachine store's view.

rmbolger:
> I just tested it with su rights and I your right, I get this error when trying to add to the localmachine store:
> > MethodInvocationException: Exception calling "Open" with "1" argument(s): "Unix LocalMachine X509Stores are read-only for all users."

The irony of that error is that it implies you can open the store as ReadOnly, but you get a similar error when you do.

> MethodInvocationException: Exception calling "Open" with "1" argument(s): "Unix LocalMachine X509Store is limited to the Root and CertificateAuthority stores."

Basically, there don't seem to be any combination of OpenFlags that succeed in opening the LocalMachine\My store on Linux.

In any case, I'll futz with it and we should be able to at least get the CurrentUser store working.

webprofusion-chrisc:
Yeah I believe it's just because Linux has no _unified_ concept of a machine store, and in dotnet for linux there is just a CurrentUser store (under ~/.dotnet I think) . macOS is different again.

https://github.com/dotnet/runtime/blob/c9cc904bc3d907333a9675168d9868d688aaca68/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/X509Certificates/StorePal.OpenSsl.cs#L141

You could perhaps ping the real expert, bartonjs , if you need the back story but there's probably a closed issue for it somewhere on the dotnet repo.

rmbolger:
It doesn't really matter for the purposes of this PR, but my other educational curiosity revolves around how the PFXs are being stored on disk on Linux when you use X509Store to add them. Like you pass a [X509Certificate2](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificate2) object to the [X509Store.Add()](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509store.add) method. But then what happens under the hood? Clearly it generates a PFX. But what password does it use? And where is that password stored since you don't need to provide it when you later need to read back in the cert via [X509Store.Certificates.Find()](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificate2collection.find)

So many questions and no desire to wade through the source to find the answers. heh

rmbolger:
Ok, cleaned everything up so we just have the X509Store implementation now and added some additional error handling and debug logging. I think this is good to go. Works on my old'ish Ubuntu system and Windows 10 workstation. But I'm going to likely give it more of a workout soon as I look into #599.

PR #605, opened 2025-03-14
Please add ACTALIS to the well-known servers (shortcuts)

defacto64:
I would like to ask that Actalis' ACME Server be added to the list of "shortcuts" (i.e. well-known servers) than can be used with Posh-ACME to specify/configure the desired ACME server.

The proposed `ACTALIS_PROD` shortcut represents the Actalis' production ACME server, that is https://acme-api.actalis.com/acme/directory

Actalis (https://actalis.com) offers, to date, up to 4 free 90-days DV certificates via ACME within our Free plan (an unlimited number is granted on subscription of any payed plan)

Actalis' ACME service _requires_ External Account Binding. Anybody can get their own EAB credentials by registering on the Actalis website and then going to the ACME management section.

I remain available for any further information.

### DIFF
--- a/Posh-ACME/Posh-ACME.psm1
+++ b/Posh-ACME/Posh-ACME.psm1
@@ -49,6 +49,7 @@ $script:WellKnownDirs = @{
     GOOGLE_STAGE = 'https://dv.acme-v02.test-api.pki.goog/directory'
     SSLCOM_RSA = 'https://acme.ssl.com/sslcom-dv-rsa'
     SSLCOM_ECC = 'https://acme.ssl.com/sslcom-dv-ecc'
+    ACTALIS_PROD = 'https://acme-api.actalis.com/acme/directory'
 }
 $script:HEADER_NONCE = 'Replay-Nonce'
 $script:USER_AGENT = "Posh-ACME/4.28.0 PowerShell/$($PSVersionTable.PSVersion)"
--- a/docs/Tutorial/index.md
+++ b/docs/Tutorial/index.md
@@ -14,7 +14,7 @@ Set-PAServer LE_STAGE
 ```
 
 !!! note
-    `LE_STAGE` is a shortcut for the Let's Encrypt Staging server's directory URL. You could do the same thing by specifying the actual URL which is `https://acme-staging-v02.api.letsencrypt.org/directory` and this module should work with any ACME compliant directory URL. Other currently supported shortcuts include `LE_PROD`, `BUYPASS_PROD`, `BUYPASS_TEST`, and `ZEROSSL_PROD`.
+    `LE_STAGE` is a shortcut for the Let's Encrypt Staging server's directory URL. You could do the same thing by specifying the actual URL which is `https://acme-staging-v02.api.letsencrypt.org/directory` and this module should work with any ACME compliant directory URL. Other currently supported shortcuts include `LE_PROD`, `BUYPASS_PROD`, `BUYPASS_TEST`, `ZEROSSL_PROD`, and `ACTALIS_PROD`.
 
 Once you set a server, the module will continue to perform future actions against that server until you change it with another call to `Set-PAServer`. The first time you connect to a server, a link to its Terms of Service will be displayed. You should review it before continuing.
 


rmbolger:
Hi @defacto64. Always happy to add more free ACME CAs to the list. Are you an Actalis employee or just a happy user?

I'd love to get some more technical details so I can update the ACME CA comparison page also maintained in this repo in `docs/Guides/ACME-CA-Comparison.md`.
https://poshac.me/docs/latest/Guides/ACME-CA-Comparison/

I've actually been meaning to add an "Operating Country" field in the CA info. The 4 cert limitation also means I'd need to add a "Free Cert Limit" field as well. I'm curious about the technical implementation around how that 4 cert limit is applied though.

- Is it 4 certs ever or just 4 currently active/non-expired certs?
- How do renewals affect the limit if the old cert is not yet expired?
- What sort of error is returned (and when) if you've exceeded the limit?
- Does the "single domain" limitation allow for an apex + www SAN like ssl.com does? Or do folks have to choose between apex and www records?

defacto64:
Thank you for replying.

I do work for Actalis, and you can also reach me at adriano.santoni@staff.aruba.it

Find below the answers to your questions:

> Is it 4 certs ever or just 4 currently active/non-expired certs?

For now, it is 4 certs ever.

> How do renewals affect the limit if the old cert is not yet expired?

For now, whether it is a renewal or not makes no difference.

> What sort of error is returned (and when) if you've exceeded the limit?

We return "Not enough quota."

> Does the "single domain" limitation allow for an apex + www SAN like ssl.com does? Or do folks have to choose between apex and www records?

For now, people have to choose; but we intend to implement a more flexible logic shortly.

rmbolger:
Ouch. 4 certs ever is rough especially when they can only have a single SAN value. I sort of intended the well known providers to only include truly free ACME CAs. But the 4 cert limit is more like a free-trial ACME CA. Is there any timeline for when "For now" changes to something users can rely on long term for at least a few sites? The current limits practically make it unusable even as a backup CA for most use-cases.

defacto64:
Could you clarify what kind of policy would make the Actalis ACME service "usable" from your point of view? 
A higher number of free certificates? Not counting renewals? Not counting expired certificates?

rmbolger:
Bare minimum is that a person/organization with a domain name should be able to get a certificate that contains at least the apex and "www" subdomain as SANs (e.g. `example.com` and `www.example.com`). They should be able to obtain and renew that certificate for free forever (or until the CA no longer exists).

But that single certificate would ultimately end up as 2 active/unexpired certificates during the window of time where the ACME client has recently requested a renewed copy of the cert, but the old copy hasn't expired yet. So any CA imposed "active" cert limits should reflect that operational reality.

So if the CA wants to limit free certs to 4 logical certs, the backend should realistically allow for 8 unless it has internal methods to recognize renewed certs and only count the most recent copy against the limit.

webprofusion-chrisc:
Jumping in, for what it's worth I agree with @rmbolger that if a CA gets included in lists like this then it needs to be clear where the "gotcha" is, e.g. where/when does the commercial element kick in. Including a commercial entity can imply a commercial relationship where none exists.

Regarding a balance of fee vs commercial usage, if you offer 4 free certs then that would _not_ be expected to count renewals, because it's the same certs covering the same names under the same ACME account. You can have rate limits as to frequency of renewal/duplication but I would not expect to reach my 4th renewal of a single certificate and suddenly need to start paying - if that's the case then they are in fact a free _trial_, and that would need to be clear.

That said, for certifytheweb.com we will probably add Acatalis as a built in CA preset, simply because we are a freemium/commercial app ourselves and we already include other commercial/enterprise presets CAs, so context is relevant, and we are starting to see a trend of European companies requiring non-US based CAs.

defacto64:
Thank you, I understand your points.
We will take your comments into consideration.
When we make changes to our policy, I will update you here.

defacto64:
@webprofusion-chrisc, please note that we are only talking about certificates requested _via ACME_.
For certificates requested in other ways (e.g., interactively) we provide up to 4 certs within the "Free" plan.

rmbolger:
I'll try to have a look this week. Assuming everything seems to be working, I'll also be running through some RFC855 spec compatibility tests so I can add to the free ACME CA comparison guide I maintain here:
https://poshac.me/docs/latest/Guides/ACME-CA-Comparison

defacto64:
@rmbolger , did you have a chance to do some testing?

defacto64:
We will look into this.

defacto64:
Thank you.
Btw, please note that BuyPass (also included in the list of well-known servers) has exited the SSL business:
https://www.buypass.com/products/tls-ssl-certificates

Issue #622, opened 2025-06-15
DNS Plugin TransIP


Tim81:
Hello Ryan,

I would appreciate a DNS Plugin for TransIP

Documentation regarding their API:
https://api.transip.nl/rest/docs.html

Thank you very much

Tim81:
If it helps, there is already a Certbot plugin

https://github.com/hsmade/certbot-dns-transip
https://packagist.org/packages/roy-bongers/certbot-transip-dns-01-validator

Tim81:
Opened Pull Request with TransIP DNS Plugin I wrote.

Tested using the Publish-Challenge and Unpublish-Challenge cmdlets and publishes/unpublished the TXT records as intended.

rmbolger:
Nice, @Tim81. Just to link the two, the PR is over in #623

rmbolger:
Plugin is live in 4.29.0

PR #623, opened 2025-06-16
Added TransIP DNS Plugin

Tim81:
Added TransIP DNS Plugin

Required parameters:
-CustomerName         is equal to your loginname at TransIP
-PrivateKey          is equal to the PrivateKey you can generate at the controlpanel of TransIP
-GlobalKey         $true means that IP Whitelisting for the used PrivateKey is turned off. $false means IP Whitelisting is in use

### DIFF
--- a/Posh-ACME/Plugins/TransIP.ps1
+++ b/Posh-ACME/Plugins/TransIP.ps1
@@ -1,0 +1,243 @@
+<#
+.SYNOPSIS
+    Posh-ACME DNS plugin for TransIP
+
+.DESCRIPTION
+    Implements DNS-01 for TransIP.
+#>
+
+# Returns the plugin type for Posh-ACME (required interface)
+function Get-CurrentPluginType { 'dns-01' }
+
+function Get-TransIPJwtToken {
+    param(
+        [Parameter(Mandatory)][string]$CustomerName,
+        [Parameter(Mandatory)][string]$PrivateKey,
+        [Parameter(Mandatory)][bool]$GlobalKey,
+        [Parameter()][string]$ApiEndpoint = "https://api.transip.nl/v6"
+    )
+    # 1. Compose payload required by TransIP
+    $nonce = -join ((65..90) + (97..122) + (48..57) | Get-Random -Count 16 | % {[char]$_})
+    $randomNum = Get-Random -Minimum 10000 -Maximum 99999
+    $label = "Posh-ACME-$randomNum"
+    $tokenBody = @{
+        login           = $CustomerName
+        nonce           = $nonce
+        global_key      = $GlobalKey
+        expiration_time = "5 minutes"
+        label           = $label
+    } | ConvertTo-Json -Compress
+    # 2. Sign the JSON body with SHA512/PKCS1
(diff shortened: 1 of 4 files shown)


dapkor (commented):
Sweet, Love to see this automated !

rmbolger:
Addresses #622

rmbolger:
It looks like this plugin will currently only work on PowerShell 7+ because [ImportFromPem](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.rsa.importfrompem) doesn't exist in .NET Framework that old PowerShell 5.1 uses.

However, it should be possible to use some of the module's own private functions to import in a way that's compatible with both. Specifically, `Import-Pem` will load the key into a BouncyCastle AsymmetricCipherKeyPair object which can then be passed to `ConvertFrom-BCKey` to get back a .NET RSACryptoServiceProvider object that you should be able to use the same way with `SignData()` as you are now.

I'd also suggest we allow for a couple options to specify the private key. Passing in as a string is fine, but we should make it a SecureString so it doesn't get saved to disk in the clear. But some folks would also probably prefer an option that let's them just pass a file path to the key file.

Tim81:
I’ll look into it, but with testing I actually used a file as source as well, but used get-content to pass it to the variable.

Tim81:
@rmbolger I've added SecureString, Filepath parameters for the PrivateKey, removed the plaintext parameter. Also added Windows Powershell 5.1 compatability.

Tim81:
Added support to supply the JWT access token directly.

Tim81:
Hi @rmbolger when will the PR reviewed further? :-)

Tim81:
I've created an account and purchased a throwaway domain to test with.
I have emailed you the credentials

Tim81:
Created different parametersets for PrivateKey and changed parametername CustomerName to TIPCustomerName, as well made the DNS $RecordName and $TxtValue the first parameters, altough due to explicit position value, it shouldn't matter.

rmbolger:
Pushed some changes that mostly revolve around parameter naming and debug logging. I need to head to bed, but I stumbled on a potential bug when I was testing adding records to the zone apex. TransIP throws an error if you try to add a record that has an existing record with the same relative name and type if the existing record's TTL doesn't match the new record's TTL.

For example, the default SPF TXT record on the domains has a TTL of 1 hour. But if the plugin tries to add another `@` record with a challenge token value and TTL of only 5 minutes, it fails. We probably need to tweak the code that checks for an existing record and grab the TTL value to re-use on the new record if one already exists with the same name+type.

Tim81:
Alright, so before actually adding a record check if one exists and use it’s TTL instead of 5 minutes. Also means that we need to get the full record before removing as the API uses the TTL value there as well. Is a required value in the API to remove a TXT record, I don’t understand why if you specify the full name and it’s value, but the API demands it.

rmbolger:
I was actually just about to test a fix for this. The API doesn't seem to care if the TTL matches on the delete request body at least if I recall correctly from last night.

rmbolger:
I was just about to push a release when you submitted the PR. So likely very soon.

PR #636, opened 2025-08-15
Add AZArcAgentAPIVersion param for Azure IMDS parameter set

semics-tech:
For legacy systems where the Azure Managed Identity Agent is not updated, we need a way to override the default api version of the


rmbolger:
Hey @semics-tech. Thanks for the contribution. I had no clue the agent version affected the IMDS url, but it makes sense. The code as submitted doesn't work, but I can see the intent. I should be able to flesh it out into a working additional plugin param.

Do you of any online reference to find what agent versions support what API versions? If it exists, I'd like to add it to the documentation as well.

PR #637, opened 2025-08-21
Update Windows.ps1

jmpederson1:
I get the following error when renewing a Windows DNS cert:

"cmdlet New-CimSessionOption at command pipeline position 1 Supply values for the following parameters:
Protocol:"

Entering 'Default' gets past this prompt. I believe the issue started in v4.29.0 with the changes to the 'New-CimSession'.

Setting $sessionOpts = @{Protocol='Default'} when declaring sessionOpts array fixes the issue for me.


jmpederson1 (commented):
I added the IF statement and removed Protocol='Default' from $sessionOpts.

This fixed my issue connecting to a Windows DNS server.

PR #638, opened 2025-08-22
Add PleskDNS plugin

felixwiehl:
This adds a DNS plugin for [Plesk](https://www.plesk.com/), a self-hosted hosting management software.


felixwiehl:
Hi @rmbolger 
I know that testing this might be quite time-consuming, as you'll need to setup plesk first to even get going.
Unfortunately i cannot provide a dev environment, but if there is something i can help you with/test, just let me know.

PR #653, opened 2025-12-26
Technitium DNS Plugin

shalafi99:
## New Plugin: Technitium DNS Server

Adds support for Technitium DNS Server via its HTTP/HTTPS API.

### Features
- Supports both HTTP and HTTPS protocols
- Configurable TTL for TXT records
- Compatible with Technitium DNS v13 and v14 (with or without clustering)
- Follows standard PoshACME plugin conventions
- Secure credential handling via SecureString

### Testing
Tested with Technitium DNS Server v14.2

**Test Results:**
- ✅ Basic add/remove operations
- ✅ Multiple TXT values (wildcard certificate support)
- ✅ Zone matching (apex and sub-zones)
- ✅ Protocol support (HTTP and HTTPS)
- ✅ Error handling (invalid credentials, network errors, permission errors)
- ✅ Custom TTL parameter
- ✅ End-to-end certificate request (staging environment)

All tests passed successfully.

### Documentation
- Comprehensive usage guide included
- Security best practices documented
- Examples for common scenarios
- Clustering considerations (v14+)

### DIFF
--- a/Posh-ACME/Private/Import-PluginDetail.ps1
+++ b/Posh-ACME/Private/Import-PluginDetail.ps1
@@ -83,6 +83,7 @@ function Import-PluginDetail {
         'SimplyCom'            = [pscustomobject]@{PSTypeName = 'PoshACME.PAPluginDetail'; ChallengeType = 'dns-01'; Path = ''; Name = 'SimplyCom'}
         'SOLIDServer'          = [pscustomobject]@{PSTypeName = 'PoshACME.PAPluginDetail'; ChallengeType = 'dns-01'; Path = ''; Name = 'SOLIDServer'}
         'SSHProxy'             = [pscustomobject]@{PSTypeName = 'PoshACME.PAPluginDetail'; ChallengeType = 'dns-01'; Path = ''; Name = 'SSHProxy'}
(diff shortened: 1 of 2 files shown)


rmbolger:
So installing a technitium test box was way easier than expected. Hooray for well maintained open source software.

One minor documentation issue I found so far. The limited user creation steps in the guide mention adding `View` and `Modify` permissions. But I was getting access denied errors with record cleanup actions until I also added `Delete` permissions.

rmbolger:
Fixed the documentation issue and added a `TechnitiumIgnoreCert` switch that can be used to disable cert validation for instances using self-signed or otherwise untrusted certs. Also tweaked the debug logs a bit so it shows a JSON representation of the body with the token redacted.

Basic functionality seems solid. But I found an edge case bug where if you try to run Publish-Challenge with the same record details as a record that already exists, you get an error. Ideally, it should either gracefully handle and ignore the error or check whether the record exists before trying to add it. Same sort of thing happens if you try to delete a record that no longer exists.

rmbolger:
The nested calls to the cert ignore functions ended up causing a bug that prematurely stopped ignoring validation before the main action api call was finished. So I fixed that and ended up refactoring all the API calls into a `Invoke-Technitium` helper function which tidied up everything nicely.

This seems to be working as expected on my test box. Can you check to make sure I haven't broken any functionality in your environment? If so, I think it's ready to merge.

PR #685, opened 2026-07-20
Establish security policy and reporting guidelines in SECURITY.md

turkiabdulazizssa-commits:
Added sections for supported versions and vulnerability reporting.


rmbolger:
Hey @turkiabdulazizssa-commits, can you provide a little more context for this PR and why you think it's necessary? It looks like a generic template that is not specific for this project.

Issue #589, opened 2024-11-21
RegRu Plugin.  Only HTTP method POST allowed


rozhkovets:
when I try use RegRu Plugin, I get an error.
The password for accessing the API is set, the IP address is added to the allowed ones.

`$pArgs = @{
    RegRuCredential = (Get-Credential)
	}`

`New-PACertificate '###' -Plugin Regru -PluginArgs $pArgs -DNSSleep 600 -AcceptTOS -Contact '###@gmail.com'`

```
Submit-ChallengeValidation : Reg.ru API call failed querying zone records: Only HTTP method POST allowed
C:\Users\NAS\Documents\WindowsPowerShell\Modules\Posh-ACME\Public\New-PACertificate.ps1:258 знак:9
+         Submit-ChallengeValidation                                                                                                          
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~                                                                                                              
+ CategoryInfo          : OperationStopped: (Reg.ru API call...od POST allowed:String) [Submit-ChallengeValidation], RuntimeException  
+ FullyQualifiedErrorId : Reg.ru API call failed querying zone records: Only HTTP method POST allowed,Submit-ChallengeValidation
```

PR #687, opened 2026-08-09
Regru: switch API calls from GET to POST with form-encoded body

kropachev:
REG.RU API 2 rejects the current plugin behavior in two ways:

1. GET on zone/* endpoints returns: { "error_code": "ONLY_POST_ALLOWED", "error_text": "Only HTTP method POST allowed" }

2. POST with params in URL query string returns: { "error_code": "QUERY_PARAMS_DISALLOWED", "error_text": "Query string parameters are disallowed. Pass parameters in form data." }

Only POST with application/x-www-form-urlencoded body is accepted.

This change moves input_format/input_data from the URL query string into a POST body via Invoke-RestMethod -Method Post -Body @{...}. input_data is passed as raw JSON produced by ConvertTo-Json (not URL-escaped) - Invoke-RestMethod handles form-encoding automatically when given a hashtable in -Body, so wrapping the JSON in [uri]::EscapeDataString would cause double-encoding.

Also fixes the Add-DnsTxt SYNOPSIS/DESCRIPTION that was copy-pasted from Remove-DnsTxt.

Reference: the official acme.sh dns_regru.sh plugin uses POST with form-body for the same endpoints and is known-working.

Refs #589

### DIFF
--- a/Posh-ACME/Plugins/Regru.ps1
+++ b/Posh-ACME/Plugins/Regru.ps1
@@ -44,13 +44,20 @@ function Add-DnsTxt {
             text                = $TxtValue
             output_content_type = 'plain'
         }
-        $queryUri = '{0}add_txt?input_format=json&input_data={1}' -f $apiBase,(ConvertTo-RrBody $bodyParams)
+        $queryUri = '{0}add_txt' -f $apiBase
+        $formBody = @{
+            input_format = 'json'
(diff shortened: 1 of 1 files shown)


rmbolger:
Thanks for the fix @kropachev. This looks like a straightforward fix and I'll take your word for it that it works.

PR #689, opened 2026-08-17
Add InfomaniakV2 plugin for the Infomaniak v2 API

Sundypha:
Adds an `InfomaniakV2` plugin and guide for [Infomaniak](https://www.infomaniak.com)'s v2 API. I wrote the original Infomaniak plugin back in #309.

### Why a separate plugin instead of updating the existing one

The v2 endpoints need a token with `dns:read` and `dns:write`, and a token carrying only the `Domain` scope that the current guide asks for is rejected outright. Changing the existing plugin in place would break existing users' tokens at their next renewal, so I followed the GoDaddy/GoDaddyV3 pattern instead. Existing orders stay on the v1 plugin until the user explicitly migrates, and the guide documents how.

Happy to fold it into the existing plugin instead if you would rather. That is the main thing I would like your call on.

### Implementation notes

v2 addresses zones by name, so the v1 `/1/product` lookup and its numeric zone ID bookkeeping are gone. Zone detection uses `/2/zones/{zone}/exists`, which answers 404 for a zone outside the account rather than 200 with `false`, so the walk treats 404 as "keep looking".

`/2/zones/{zone}/exists` also returns `false` for a domain that is registered with Infomaniak but whose nameservers point elsewhere. The API still exposes a leftover zone full of old records in that case, so testing existence by listing records would happily write challenge records into a zone nobody queries. The guide warns about this.

### Testing

- Full challenge cycle (create, duplicate detection, delete, idempotent delete) against two separate Infomaniak hosted zones in the same account, using a token carrying only `dns:read` and `dns:write`. Both zones came back byte identical afterwards.
- Those two scopes are therefore necessary and sufficient. `domain:read` is not needed, because the plugin probes candidate zones by name rather than enumerating the account's domains.
- A token carrying only the legacy `Domain` scope is rejected by every v2 endpoint. A v2 scoped token also works against the v1 endpoints, so a single token covers both plugins.
- Zone walk verified at one, two, three and four labels below the zone apex, including that the API accepts and round trips multi label `source` values like `_acme-challenge.a.b`.
- End to end against Let's Encrypt staging: issued with `-Plugin Infomaniak`, migrated the order with `Set-PAOrder -Plugin InfomaniakV2`, then `Submit-Renewal` reissued the certificate with nothing left in the zone. Both plugins use the `InfomaniakToken` parameter name, so the stored plugin args carry across the switch unchanged, as long as the token already carries the v2 scopes. The whole sequence ran on a single `dns:read` plus `dns:write` token.
- `Invoke-Pester`: 301 passed, 0 failed on both Windows PowerShell 5.1 and PowerShell 7.4.2.

### Testing on your side

I know you prefer to test plugins yourself, and Infomaniak has no free tier or sandbox. The account I tested against hosts live business domains, so I would rather not hand out credentials to it.

What I can offer instead: run whatever test scenarios you want against my account and paste the full output, including any edge cases you want covered before merging.

### DIFF
--- a/Posh-ACME/Plugins/InfomaniakV2.ps1
+++ b/Posh-ACME/Plugins/InfomaniakV2.ps1
@@ -1,0 +1,303 @@
+function Get-CurrentPluginType { 'dns-01' }
+
+function Add-DnsTxt {
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory, Position = 0)]
+        [string]$RecordName,
+        [Parameter(Mandatory, Position = 1)]
+        [string]$TxtValue,
+        [Parameter(Mandatory, Position = 2)]
+        [securestring]$InfomaniakToken,
+        [Parameter(ValueFromRemainingArguments)]
+        $ExtraParams
+    )
+
+    $apiRoot = 'https://api.infomaniak.com/2/zones'
+
+    # grab the plain text token
+    $tokenPlain = [pscredential]::new('a',$InfomaniakToken).GetNetworkCredential().Password
+
+    # build the common parameters for all API calls
+    $commonParams = @{
+        Headers = @{
+            Authorization = "Bearer $tokenPlain"
+            Accept        = 'application/json'
+        }
+        ErrorAction = 'Stop'
+        Verbose = $false
+        Debug = $false
+    } + $script:UseBasic
+
+    if (-not ($zone = Find-IKV2Zone $RecordName $commonParams $apiRoot)) {
+        throw "Unable to find matching zone for $RecordName."
+    }
+
+    # separate the portion of the name that doesn't contain the zone name
+    $recShort = $RecordName -ireplace "\.?$([regex]::Escape($zone.TrimEnd('.')))$",''
+    # the API represents an apex record with a bare dot
+    if ([String]::IsNullOrEmpty($recShort)) { $recShort = '.' }
+
+    if (Find-IKV2TxtRecord $zone $recShort $RecordName $TxtValue $commonParams $apiRoot) {
+        Write-Debug "Record $RecordName already contains $TxtValue. Nothing to do."
+        return
+    }
+
+    try {
+        Write-Verbose "Adding a TXT record for $RecordName with value $TxtValue"
+        $queryParams = @{
+            Uri = "$apiRoot/$zone/records"
+            Method = 'POST'
+            Body = @{
+                type   = 'TXT'
+                source = $recShort
+                target = $TxtValue
+                ttl    = 600
+            } | ConvertTo-Json -Compress
+            ContentType = 'application/json'
+        } + $commonParams
+        Write-Debug "POST $($queryParams.Uri)`n$($queryParams.Body)"
+        Invoke-RestMethod @queryParams | Out-Null
+    } catch { throw }
+
+    <#
+    .SYNOPSIS
+        Add a DNS TXT record to Infomaniak using their v2 API.
+
+    .DESCRIPTION
+        Add a DNS TXT record to Infomaniak using their v2 API.
+
+    .PARAMETER RecordName
+        The fully qualified name of the TXT record.
+
+    .PARAMETER TxtValue
+        The value of the TXT record.
+
+    .PARAMETER InfomaniakToken
+        The API token for your Infomaniak account. It must have the 'dns:read' and 'dns:write' scopes.
+
+    .PARAMETER ExtraParams
+        This parameter can be ignored and is only used to prevent errors when splatting with more parameters than this function supports.
+
+    .EXAMPLE
+        $token = Read-Host 'Token' -AsSecureString
+        Add-DnsTxt '_acme-challenge.example.com' 'txt-value' $token
+
+        Adds the specified TXT record with the specified value.
+    #>
+}
+
+function Remove-DnsTxt {
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory, Position = 0)]
+        [string]$RecordName,
+        [Parameter(Mandatory, Position = 1)]
+        [string]$TxtValue,
+        [Parameter(Mandatory, Position = 2)]
+        [securestring]$InfomaniakToken,
+        [Parameter(ValueFromRemainingArguments)]
+        $ExtraParams
+    )
+
+    $apiRoot = 'https://api.infomaniak.com/2/zones'
+
+    # grab the plain text token
+    $tokenPlain = [pscredential]::new('a',$InfomaniakToken).GetNetworkCredential().Password
+
+    # build the common parameters for all API calls
+    $commonParams = @{
+        Headers = @{
+            Authorization = "Bearer $tokenPlain"
+            Accept        = 'application/json'
+        }
+        ErrorAction = 'Stop'
+        Verbose = $false
+        Debug = $false
+    } + $script:UseBasic
+
+    if (-not ($zone = Find-IKV2Zone $RecordName $commonParams $apiRoot)) {
+        throw "Unable to find matching zone for $RecordName."
+    }
+
+    # separate the portion of the name that doesn't contain the zone name
+    $recShort = $RecordName -ireplace "\.?$([regex]::Escape($zone.TrimEnd('.')))$",''
+    # the API represents an apex record with a bare dot
+    if ([String]::IsNullOrEmpty($recShort)) { $recShort = '.' }
+
+    $rec = Find-IKV2TxtRecord $zone $recShort $RecordName $TxtValue $commonParams $apiRoot
+
+    if (-not $rec) {
+        Write-Debug "Could not find record $RecordName to delete. Nothing to do."
+        return
+    }
+
+    try {
+        Write-Verbose "Removing TXT record for $RecordName with value $TxtValue"
+        $queryParams = @{
+            Uri = "$apiRoot/$zone/records/$($rec.id)"
+            Method = 'DELETE'
+        } + $commonParams
+        Write-Debug "DELETE $($queryParams.Uri)"
+        Invoke-RestMethod @queryParams | Out-Null
+    } catch { throw }
+
+    <#
+    .SYNOPSIS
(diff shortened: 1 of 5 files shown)


rmbolger:
This is fantastic work, @Sundypha. Thank you! I'm actually ok not testing this one myself. But can you make sure you test the standard publish commands using a zone apex with `-DnsAlias` if you haven't already.

Also more importantly, test the new `Publish-DnsPersistChallenge` function. That one has been causing some strife in other plugins because the `$TxtValue` string comes in with embedded double quotes surrounding the value due to the included space characters. In retrospect, I wish I would have had the original dns-01 challenge values sent that way as well, but hindsight is 20/20.

These are the rough commands I usually use to test (at least for the current draft-01 implementation which is still current). The content of AccountUri and IssuerDomainName shouldn't matter as long as they're non-empty. Mental note, add this to the plugin dev guide.

```powershell
$pubParams = @{
    Domain = 'example.com'
    AccountUri = 'fakeaccount'
    IssuerDomainName 'fakeissuer'
    AllowWildcard = $true
    Plugin = 'InfomaniakV2'
    PluginArgs = $pArgs
    Verbose = $true
}
Publish-DnsPersistChallenge @pubParams
Unpublish-DnsPersistChallenge @pubParams
```

If it works as-is, great! If not, what I've been doing in some of the other plugins that needed fixing is something like below to basically normalize the quotes on the value before starting to work with it. Then you can continue with the rest of the API stuff knowing they're already in place.

```powershell
# Normalize the TxtValue to ensure it is wrapped in quotes
if ($TxtValue -notmatch '^".*"$') {
    $TxtValue = "`"$TxtValue`""
}
```

Sundypha:
I'm sorry for this lengthy response. I hope with showing you the test-cases I ran we didn't left a hole anywhere. If you find something odd let me know. 

Your instinct was right, `Publish-DnsPersistChallenge` was broken. Your commands against the old code:

```
publish #1    ok
publish #2    THREW: 400 Bad Request (dns_record_already_exists)
unpublish     ok          <- reported success, removed nothing
records left  1           LEAKED "fakeissuer; accounturi=fakeaccount; policy=wildcard"
```

After the fix: all three ok, 0 records left.

I used your normalization snippet verbatim in `Add-DnsTxt` and `Remove-DnsTxt` and deleted my own quote-stripping helper. It fits this API well, because Infomaniak stores TXT targets as exactly one pair of quotes regardless of input, so normalizing on the way in lets me compare against the stored target as-is:

| submitted | stored |
|---|---|
| `abc` | `"abc"` |
| `"abc"` | `"abc"` |
| `"fakeissuer; accounturi=fakeaccount; policy=wildcard"` | identical, one pair |

Only deviation: written as `'"{0}"' -f $TxtValue` to match how `Publish-DnsPersistChallenge` builds the value. Happy to switch to the backtick form.

I then worked through the Testing Plugins section of the plugin development guide, including the `test1`/`test2`/`test3` multiple-values pattern and the `-DnsAlias` apex example. Everything below ran against two real Infomaniak hosted zones, shown here as `example.com` and `example.net`, using a token carrying only `dns:read` and `dns:write`.

```powershell
$pArgs = @{ InfomaniakToken = (ConvertTo-SecureString $token -AsPlainText -Force) }
```

### dns-persist-01

Standalone, per your snippet:

```powershell
$pp = @{
    Domain           = 'example.com'
    AccountUri       = 'fakeaccount'
    IssuerDomainName = 'fakeissuer'
    AllowWildcard    = $true
    Plugin           = 'InfomaniakV2'
    PluginArgs       = $pArgs
}
Publish-DnsPersistChallenge   @pp     # writes _validation-persist.example.com
Publish-DnsPersistChallenge   @pp     # again, must match not duplicate
Unpublish-DnsPersistChallenge @pp
```

And against a real LE staging order, which also exercises the challenge parsing and the account URI fallback:

```powershell
Set-PAServer LE_STAGE
New-PAAccount -AcceptTOS
New-PAOrder -Domain 'pa-persist.example.com' -Plugin InfomaniakV2 -PluginArgs $pArgs -Force
$order = Get-PAOrder 'pa-persist.example.com'
Publish-DnsPersistChallenge   -Order $order -AllowWildcard
Publish-DnsPersistChallenge   -Order $order -AllowWildcard
Unpublish-DnsPersistChallenge -Order $order -AllowWildcard
```

| Step | Result | Notes |
|---|---|---|
| standalone publish | pass | one quote pair, value stored exactly as passed |
| standalone re-publish | pass | before the fix: 400 `dns_record_already_exists` |
| standalone unpublish | pass | before the fix: returned success, removed nothing |
| standalone with no `Set-PAServer` and no account | pass | that parameter set needs neither |
| `-Order` publish against LE staging | pass | wrote `"letsencrypt.org; accounturi=https://acme-staging-v02.api.letsencrypt.org/acme/acct/<id>; policy=wildcard"` |
| `-Order` re-publish | pass | matched, no duplicate |
| `-Order` unpublish | pass | removed, zero residue |

The `-Order` run went through a real LE staging order, so the value came from the live challenge rather than my placeholders. Staging omits `accounturi`, so the account URI came from the fallback, and the record ended up carrying the real one. That makes it a better quoting test than my synthetic value, since it contains `:`, `/`, `=`, `;` and spaces and still stored as a single quote pair.

I did not attempt full issuance. The challenge type comes from the plugin's `ChallengeType`, overridable only via `$Order.DnsVariant`, which is currently `ValidateSet('dns-01','dns-account-01')`, so dns-persist-01 cannot be selected for validation yet.

### Zone apex via -DnsAlias

```powershell
$publishParams = @{
    Domain     = 'my.cname.tld'
    Account    = (Get-PAAccount)
    Token      = 'fake-token'
    Plugin     = 'InfomaniakV2'
    PluginArgs = $pArgs
    DnsAlias   = 'example.com'
}
Publish-Challenge   @publishParams
Unpublish-Challenge @publishParams
```

| Step | Result | Notes |
|---|---|---|
| publish to apex | pass | record created at the apex |
| unpublish from apex | pass | removed, apex left with only its pre-existing records |
| two values at the apex, removed one at a time | pass | removing the first left the second intact |
| raw API apex encodings `"."`, `""`, omitted | pass | all three canonicalize to `source: "."`, so this provider wants `.` rather than `@` |

### Standard dns-01

```powershell
Publish-Challenge   example.com (Get-PAAccount) test1 InfomaniakV2 $pArgs
Publish-Challenge   example.com (Get-PAAccount) test2 InfomaniakV2 $pArgs
Publish-Challenge   example.com (Get-PAAccount) test3 InfomaniakV2 $pArgs
Unpublish-Challenge example.com (Get-PAAccount) test2 InfomaniakV2 $pArgs
Unpublish-Challenge example.com (Get-PAAccount) test1 InfomaniakV2 $pArgs
Unpublish-Challenge example.com (Get-PAAccount) test3 InfomaniakV2 $pArgs
```

| Step | Result | Notes |
|---|---|---|
| single publish then unpublish | pass | 1 record then 0 |
| three values on one record name | pass | all three coexist as separate records |
| removing only the middle value | pass | `test1` and `test3` survived, matching is on name and value |
| removing the remaining two | pass | zero residue |

Plugin functions called directly, each as add, add again, remove, remove again:

```powershell
Add-DnsTxt '_pa-d1.example.com'       $val $sec     # source _pa-d1
Add-DnsTxt '_pa-d2.a.example.com'     $val $sec     # source _pa-d2.a
Add-DnsTxt '_pa-d3.a.b.example.com'   $val $sec     # source _pa-d3.a.b
Add-DnsTxt '_pa-d4.a.b.c.example.com' $val $sec     # source _pa-d4.a.b.c
Add-DnsTxt '_pa-zoneb.example.net'    $val $sec     # second zone
Add-DnsTxt 'example.com'              $val $sec     # apex, source .
```

| Record | Result | Notes |
|---|---|---|
| 1 label below apex | pass | |
| 2 labels below apex | pass | |
| 3 labels below apex | pass | |
| 4 labels below apex | pass | multi label `source` accepted and round tripped |
| second zone, same account | pass | |
| zone apex | pass | |

### End to end against Let's Encrypt staging

```powershell
New-PACertificate 'pa-e2e.example.com' -Plugin Infomaniak   -PluginArgs $pArgs -DnsSleep 45
Set-PAOrder       'pa-e2e.example.com' -Plugin InfomaniakV2 -PluginArgs $pArgs
Submit-Renewal    'pa-e2e.example.com' -Force
```

| Step | Result | Notes |
|---|---|---|
| issue with the v1 Infomaniak plugin | pass | |
| `Set-PAOrder` to InfomaniakV2 | pass | stored PluginArgs carried over unchanged |
| `Submit-Renewal` through v2 | pass | reissued with a new thumbprint |

### Module and token scopes

```powershell
powershell.exe -C "Import-Module Pester -MinimumVersion 5.2 -MaximumVersion 5.99.99; Invoke-Pester"
pwsh          -c "Import-Module Pester -MinimumVersion 5.2 -MaximumVersion 5.99.99; Invoke-Pester"
```

| Test | Result | Notes |
|---|---|---|
| Pester, Windows PowerShell 5.1 | 301 / 0 | |
| Pester, PowerShell 7.4.2 | 301 / 0 | |
| plugin registered, discoverable, reports dns-01 | pass | |
| original Infomaniak plugin unaffected | pass | |
| token with only `dns:read` + `dns:write` | pass everywhere | `domain:read` not needed |
| token with only the legacy `Domain` scope | rejected on every v2 endpoint | 403 `all_scopes`, names the missing scope |
| `dns:read` + `dns:write` against the v1 endpoints | pass | one token serves both plugins |

### API behaviour

| Call | Result | Notes |
|---|---|---|
| `GET /2/zones/{zone}/exists` for a zone not in the account | 404 | |
| `GET /2/zones/{zone}/exists` for a domain delegated to third party nameservers | `false` | the stale zone is still listable, so a records based existence check would write into a zone nobody queries |
| `GET /1/domain/{id}/dns/record` | 200 | v1 still reachable with a v2 scoped token |
| final zone state | pass | both zones identical after every run |

rmbolger:
Perfect, @Sundypha. Merging now.

PR #690, opened 2026-09-07
Add DNSMint DNS plugin

kxbnb:
Adds a `dns-01` plugin for [DNSMint](https://dnsmint.com/).

DNSMint mints hostnames on domains it operates and serves them from its own authoritative nameservers, so a challenge is published through its API rather than a zone the user runs. The API derives the hostname from the challenge name, which means the plugin has no zone to find and no record id to track: the value published is the value removed. Wildcards work because the API keeps the two newest values for a name.

`Save-DnsTxt` is empty; there is nothing to commit.

Included:

- `Posh-ACME/Plugins/DNSMint.ps1`
- `docs/Plugins/DNSMint.md`
- entries in `Import-PluginDetail.ps1` and `docs/Plugins/index.md`

Tested on PowerShell 7.6 (Ubuntu 24.04) against Let's Encrypt staging, issuing a certificate for a wildcard plus its apex. Both challenges publish to the same record name and validate.

On the account problem in CONTRIBUTING: I work on DNSMint, and I am happy to set you up with a standing account so you can test this and support it later. Mail hello@dnsmint.com, or say the word here and I will reach out.


kxbnb:
Worked through the Testing Plugins section of the plugin development guide. PowerShell 7.6.5, Posh-ACME 4.34.0, Let's Encrypt staging, against a hostname on a domain DNSMint operates. The token carries a single `dns01:write` grant scoped to that hostname.

```powershell
$pArgs = @{ DNSMintToken = (ConvertTo-SecureString $token -AsPlainText -Force) }
```

### Standard dns-01

```powershell
Publish-Challenge   $h (Get-PAAccount) test1 DNSMint $pArgs
Publish-Challenge   $h (Get-PAAccount) test2 DNSMint $pArgs
Publish-Challenge   $h (Get-PAAccount) test3 DNSMint $pArgs
Unpublish-Challenge $h (Get-PAAccount) test2 DNSMint $pArgs
Unpublish-Challenge $h (Get-PAAccount) test1 DNSMint $pArgs
Unpublish-Challenge $h (Get-PAAccount) test3 DNSMint $pArgs
```

| Step | Result | Notes |
|---|---|---|
| single publish then unpublish | pass | 1 value then 0 |
| three values on one name | pass, with a cap | only `test2` and `test3` are served; the API keeps the two newest and `test1` was evicted |
| removing the middle value | pass | removed `test2`, `test3` survived, so matching is on name and value |
| removing the remainder | pass | zero residue |

The two-value cap is the API, not the plugin. DNSMint mirrors acme-dns here, which is what the apex plus wildcard pair needs. Worth knowing that two concurrent orders for one hostname will evict each other.

### dns-persist-01

Standalone, per the guide's snippet:

```powershell
$pp = @{
    Domain           = $h
    AccountUri       = 'fakeaccount'
    IssuerDomainName = 'fakeissuer'
    AllowWildcard    = $true
    Plugin           = 'DNSMint'
    PluginArgs       = $pArgs
}
Publish-DnsPersistChallenge   @pp
Publish-DnsPersistChallenge   @pp
Unpublish-DnsPersistChallenge @pp
```

And against a real staging order, so the value comes from the live challenge rather than placeholders:

```powershell
New-PAOrder -Domain $h -Plugin DNSMint -PluginArgs $pArgs -Force
$order = Get-PAOrder $h
Publish-DnsPersistChallenge   -Order $order -AllowWildcard
Publish-DnsPersistChallenge   -Order $order -AllowWildcard
Unpublish-DnsPersistChallenge -Order $order -AllowWildcard
```

| Step | Result | Notes |
|---|---|---|
| standalone publish | pass | `"fakeissuer; accounturi=fakeaccount; policy=wildcard"`, one quote pair |
| standalone re-publish | pass | matched, no duplicate |
| standalone unpublish | pass | zero residue |
| `-Order` publish | pass | wrote `"letsencrypt.org; accounturi=https://acme-staging-v02.api.letsencrypt.org/acme/acct/<id>; policy=wildcard"` |
| `-Order` re-publish | pass | matched, no duplicate |
| `-Order` unpublish | pass | zero residue |

On the quoting note in the guide: the value arrives wrapped and DNSMint strips the presentation quotes server-side, so the plugin does not normalise and does not need to. The `-Order` value is the better test of that, since it carries `:`, `/`, `=`, `;` and spaces and still stores as one pair.

Re-publishing an identical grant is a no-op rather than a delete and re-insert, so the record's TTL keeps ramping instead of resetting.

I did not attempt full issuance, for the same reason you'll recognise from #689: the challenge type comes from the plugin's type and is overridable only via `$Order.DnsVariant`, which is `ValidateSet('dns-01','dns-account-01')` in 4.34.0, so dns-persist-01 cannot be selected for validation yet. Confirmed on this install.

### Zone apex via -DnsAlias

```powershell
$publishParams = @{
    Domain     = 'my.cname.tld'
    Account    = (Get-PAAccount)
    Token      = 'fake-token'
    Plugin     = 'DNSMint'
    PluginArgs = $pArgs
    DnsAlias   = "_acme-challenge.$h"
}
Publish-Challenge   @publishParams
Unpublish-Challenge @publishParams
```

| Step | Result | Notes |
|---|---|---|
| publish to alias | pass | record created at the alias |
| unpublish from alias | pass | zero residue |
| alias given as a bare hostname | refused | `fqdn must be a validation name, "_acme-challenge.<hostname>." or "_validation-persist.<hostname>."` |

The refusal is the API, and it is an access rule rather than apex handling: a DNSMint account may write the two validation names under a hostname it holds and nothing else. The plugin cannot work around it and I have not tried to.

The apex condition the guide is testing for does not arise with this provider. The plugin does no zone matching and no relative-name calculation, so there is no empty short name to mishandle.

The case that matters works, and it is the acme-dns pattern: point `_acme-challenge.your-domain.com` at `_acme-challenge.<dnsmint-host>` and DNSMint answers the challenge and nothing else.

Pester passes with this branch on all four targets: https://github.com/dnsmint/Posh-ACME/actions/runs/34547360868

rmbolger:
Hey @kxbnb, thanks for the PR. I hadn't heard of DNSMint before, but it seems like an interesting service. The limitations described in your testing with apex zones make sense. It might be good to mention them in the guide though.

I'll spin up a starter account and then reach out about the standing account.

kxbnb:
Added to the guide in 137d1f9: a section on what a key may write, the `-DnsAlias` form that follows from it, and the CNAME pattern that covers a domain of your own including its apex.

On the account: a hostname on `dnsmint-108bd2.top` with a key scoped to it is set aside for you, so you do not have to spend anything to test renewals. Say the word and I will send the details wherever you prefer.

rmbolger:
Whoops. I created a starter account before I read the last comment. Will that screw things up? Should I cancel it? I sent an email to the hello address from the email I associated with the account.

kxbnb:
Don't cancel — nothing is wrong with your account. A dedicated domain is being provisioned for you now. I'll follow up on your email about the billing side.

rmbolger:
The provided API definitely makes this plugin really simple to implement which is a nice change. The only minor suggestion I'd make which I've started doing with some other plugins is to have the Invoke-RestMethod call include `-Verbose:$false` and `-Debug:$false`. This will prevent users from accidentally logging their API key on newer versions of PowerShell when they have debug output enabled. Then in lieu of the auto-debug output, I'll do a Write-Debug with the method and URI and then just the body on another line which keeps the debug intent, but also effectively back-ports it to PowerShell 5.1.

Unrelated to the plugin, I have some suggestions you might want to consider for the API itself and what it allows or doesn't allow.

Realistically, the format of the `_validation-persist` record may change pretty significantly in the next draft update. Not sure how much you want to keep track of that and update the format requirements as necessary or just loosen the rules about what is allowed. I'd guess there will end up being some overlap between CAs using the draft-01 and draft-02+ you might want to be ready for.

I believe there are also a couple CAs now allowing the use of [dns-account-01](https://datatracker.ietf.org/doc/draft-ietf-acme-dns-account-label/03/) validation which adds an account hash prefix to the standard `_acme-challenge` record. So clients would be attempting to create a record for `_<hash>._acme-challenge.<host>.<domain>`. But the API doesn't currently allow it.

kxbnb:
Thank you for the feedback, @rmbolger.

Fixed in `7b8b0fe`. `-Verbose:$false -Debug:$false` on the `Invoke-RestMethod`, and a `Write-Debug` pair in its place: method and URI, then the body. Reproduced against a local listener before and after:

```
without the switches:  token present in debug output: True
with them:             token present in debug output: False
```

**dns-account-01** is in, live today. `Get-DnsAcctLabel` and our parser agree character for character:

```
https://example.com/acme/acct/ExampleAccount  ->  _ujmmovf2vn55tgye._acme-challenge
```

and a publish/withdraw round trip through the live API:

```
name   : _jbmhwlsiorga4uwa._acme-challenge.poshacme-test.dnsmint-108bd2.top.
present: 200
stored : record at that name, holding the token
cleanup: 200, 0 records left
```

So `Publish-Challenge -DnsVariant dns-account-01` works against the plugin as it stands, standalone, the same way you tested dns-persist-01. No plugin change needed: the record name arrives already built and we forward it unchanged.

**`_validation-persist`**: unknown tags are already ignored, per draft-01 §4, so a later draft adding a parameter does not break us. A draft that changes the shape rather than adding to it would, and we have not decided how loose to be. We are tracking the draft.

Both are on https://dnsmint.com/standards under Implemented.

rmbolger:
Looks good from my end. I'm ready to merge.

